PR #58, opened 2019-10-14
Automatically deploy documentation to docs.contao.org

Toflar:
This introduces automated deployment of the master branch to docs.contao.org on every push to the master branch.

### DIFF
--- a/.github/workflows/build_and_deploy.yaml
+++ b/.github/workflows/build_and_deploy.yaml
@@ -1,0 +1,47 @@
+name: Build and deploy to https://docs.contao.org
+
+on:
+    push:
+        branches:
+            - master
+
+jobs:
+    build_and_deploy:
+        runs-on: ubuntu-18.04
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v1.0.0
+              with:
+                  submodules: true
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_deploy.yaml
```diff
@@ -0,0 +1,47 @@
+name: Build and deploy to https://docs.contao.org
+
+on:
+    push:
+        branches:
+        - master
```
leofeyer: Is this line correctly indented?

Toflar: Nope :D

Toflar: Fixed.

aschempp:
Does this include multilingual manuals? @fritzmg told me the current URL is actually wrong, as it does not contain `de`.

Toflar:
That's out of scope here. That's a build task, not a deployment task.

fritzmg:
> Does this include multilingual manuals? @fritzmg told me the current URL is actually wrong, as it does not contain `de`.

Yes, once we re-introduce the language parameter, it will be automatically deployed as well.

PR #80, opened 2019-10-15
Fixes for the introduction chapter

xchs:
BTW: The changes regarding the German lowercase umlaut "ü" are related to the macOS behavior which apparently renders and store this umlaut as precomposed characters.

_(I will review the other chapters in the next days.)_


netzarbeiter:
Thank you @xchs

PR #82, opened 2019-10-15
Correct typing error "weisst"

Mynyx:


Toflar:
Just as a hint: These are not a typing errors. They were written like that because it's the correct Swiss spelling. There's no `ß` for us :) I'm fine changing that for consistency of course but just to let you know that there will be more of those...forever 😄

netzarbeiter:
Thank you @Mynyx

PR #85, opened 2019-10-15
Update README.md

fkaminski:
Update README.md with further informatioen

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,22 @@
 
 The documentation for the Contao project will be maintained in this repository.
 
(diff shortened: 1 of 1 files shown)


fkaminski:
Sorry - have a content error here:
It should be: 
git clone --recurse-submodules https://github.com/contao/docs.git
instead of
git clone --recurse-submodules https://github.com/fkaminski/docs.git

So how do i can update/fix this now?

fkaminski:
Will try again with new PR

PR #105, opened 2019-10-20
add language parameter (for all languages)

fritzmg:
This (re-)adds the language parameters to the URLs, so that URLs will be correct in the future.

A few things to note here:

* We are using the `url` [front matter variable](https://gohugo.io/content-management/front-matter/) in order to "translate" the URL path for a page. However, this parameter overrides _the whole_ URL, including any language parameter that would be present otherwise. Thus we need to manually include that (this PR fixes that for each page).

* Originally I wanted to completely disable the English language, as long as the English translation is still incomplete. However, since there would be only one language left, Hugo automatically does not add the language prefix any more then. Thus the English language is currently enabled and I have translated at least the home page and the index page for "Extensions" - since those would be the only "visible" pages for the English page tree.

* We had some trouble with images not showing up before, but I think it is fine now.

### DIFF
--- a/docs/manual/extensions/cca-merger2.en.md
+++ b/docs/manual/extensions/cca-merger2.en.md
@@ -1,7 +1,6 @@
 ---
 title: "Merger²"
 description: "Merger² is a frontend module to display modules/contents under certain conditions."
-url: "de/extension/merger2"
 ---
 
 **[contao-community-alliance/merger2](https://packagist.org/packages/contao-community-alliance/merger2)**
--- a/docs/manual/extensions/cca-merger2.en.md
+++ b/docs/manual/extensions/cca-merger2.en.md
@@ -1,7 +1,7 @@
 ---
 title: "Merger²"
 description: "Merger² is a frontend module to display modules/contents under certain conditions."
-url: "extension/merger2"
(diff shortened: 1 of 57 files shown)


docs/manual/extensions/cca-merger2.en.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Merger²"
 description: "Merger² is a frontend module to display modules/contents under certain conditions."
-url: "extension/merger2"
+url: "de/extension/merger2"
```
netzarbeiter: This is the english version of the file.

docs/manual/extensions/metamodels.en.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: "MetaModels"
 description: "MetaModels is an extension for CMS Contao and provides a flexible and easy way to build your own data models."
-url: "extension/metamodels"
+url: "de/extension/metamodels"
```
netzarbeiter: This is the english version of the file.

netzarbeiter (changes requested):
Can you please change the url for the english files.

netzarbeiter (approved):
Thank you

PR #112, opened 2019-10-23
Polishing latest CONTRIBUTING.md

Toflar:


CONTRIBUTING.md
```diff
@@ -93,26 +95,31 @@ This should contain specific examples and use cases on how to use and implement
 
 * The paths and filenames should be in English
 * Consider that all markdown headings are taken over in the page navigation of the documentation
-* Use images by copying existing or create new ones only
```
Toflar, line 96: I had to remove this line because I did not understand what this should tell me? What was the intention here @fkaminski or @fritzmg? Did you maybe mean we should not link to external images because they might be removed?

fritzmg: Yeah, this should be made more clear. Basically the thing is: any new article should provide their own images, instead of re-using existing ones, to avoid broken images if an image is changed or moved for another article.

/cc @netzarbeiter

Toflar: Checkout 4ae2a24 then :)

fritzmg (approved):

PR #117, opened 2019-10-25
Avoid belittling words

Toflar:
I did the same PR for the Symfony docs and I think it applies to us as well (in fact I'm convinced it applies for any documentation).

### DIFF
--- a/docs/dev/framework/cron.md
+++ b/docs/dev/framework/cron.md
@@ -57,7 +57,7 @@ an associative array with the following keys, representing the available interva
 * `weekly`
 * `monthly`
 
-To register simply add another array item with the class and method
+To register your job, add another array item with the class and method
 of your cron job to one of the intervals in your `contao/config/config.php`:
 
 ```php
--- a/docs/dev/framework/cron.md
+++ b/docs/dev/framework/cron.md
@@ -57,7 +57,7 @@ an associative array with the following keys, representing the available interva
 * `weekly`
 * `monthly`
 
-To register your job, add another array item with the class and method
+To register your own job, add another array item with the class and method
 of your cron job to one of the intervals in your `contao/config/config.php`:
 
 ```php
--- a/docs/dev/getting-started/extension.md
+++ b/docs/dev/getting-started/extension.md
@@ -225,7 +225,7 @@ the extension's `composer.json`:
 ```
 
 After running `composer update`, the Contao Manager Plugin will load this bundle
-in the Contao Managed Edition. Of course, this will have no real effect yet, since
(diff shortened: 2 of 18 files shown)


fiedsch:
I like the PR but think it might be over the top in some places by "simply" removing the belittling words. 
Example: in `docs/dev/framework/caching.md` 
>  Moreover, you don't have to register to all the different callbacks such as `onsubmit_callback` or `ondelete_callback`.
>
> You can ~~simply~~ register to the [`oninvalidate_cache_tags` callback][5] and add your own tags.	You can register to the [`oninvalidate_cache_tags` callback][5] and add your own tags.

I understand the "simply" like "it suffices to ..." which I would not consider belittling, but it slightly changes the sense of the sentence.

Of course just my 2ct.

Edited to make the quotation clearer.

xchs (dismissed):

Toflar:
> I understand the "simply" like "it suffices to ..." which I would not consider belittling, but it slightly changes the sense of the sentence.

Of course it depends on the meaning. I agree. But at the same time even when used "correctly" it mostly doesn't bring any additional information to the sentence. So by removing all of them, it's easier to track down the bad ones :)

docs/dev/framework/cron.md
```diff
@@ -57,7 +57,7 @@ an associative array with the following keys, representing the available interva
 * `weekly`
 * `monthly`
 
-To register simply add another array item with the class and method
+To register your job, add another array item with the class and method
```
fritzmg: may be `your own job` here

docs/dev/getting-started/extension.md
```diff
@@ -225,7 +225,7 @@ the extension's `composer.json`:
 ```
 
 After running `composer update`, the Contao Manager Plugin will load this bundle
-in the Contao Managed Edition. Of course, this will have no real effect yet, since
+in the Contao Managed Edition. This will have no real effect yet, since
 the extension is still pretty empty.
```
fritzmg, line 229: What about `pretty` here? :) May be use `almost` instead?

fritzmg (changes requested):

fritzmg:
👍 

I actually wanted to incorporate this right from the start but unfortunately forgot about it over time and did not include it in the `CONTRIBUTING.md`.

docs/dev/getting-started/extension.md
Toflar: `pretty` sounds okay to me :)

docs/dev/framework/cron.md
Toflar: Added :)

fritzmg (approved):

PR #118, opened 2019-10-26
Explain how to register callbacks

richardhj:
Callbacks can be registered with using annotations (autowiring) or via a tagged service. I was thinking, these examples are missing.

### DIFF
--- a/docs/dev/reference/dca/callbacks.md
+++ b/docs/dev/reference/dca/callbacks.md
@@ -431,4 +431,50 @@ a button for an import "wizard".
 **return:** `string` HTML for the button
 {{% /expand %}}
 
+
+***
+
+## Examples of defining callbacks
+
+You can register any of the abovementioned callbacks by using annotations:
+
+```php
+// src/Controller/EventListener/DataContainer/MyTableListener.php
+
+namespace App\EventListener\DataContainer;
+
+use Contao\CoreBundle\ServiceAnnotation\Callback;
+use Contao\DataContainer;
+use Terminal42\ServiceAnnotationBundle\ServiceAnnotationInterface;
+
+class MyTableListener implements ServiceAnnotationInterface
+{
+
+    /**
+     * @Callback(table="tl_my_table", target="fields.xxMyField.options")
+     */
+    public function onOptionsCallback(DataContainer $dc): array
+    {
+        return ['test1', 'test2'];
+    }
+}
+```
+
+When [autowiring][autowiring] is enabled in your app, your callback is registered successfully.
+
+In case you don't use autowiring, you need to tag the corresponding service of your callback
+class:
+
+```yml
+// config/services.yml
+
+App\EventListener\DataContainer\MyTableListener:
+  tags:
+    - { name: contao.callback, table: tl_my_table, target: fields.xxMyField.options [, method: onOptionsCallback] }
+```
+
+Same as with [Hooks][hooks], the method is automatically generated if not given. As for
+example `onLoadCallback` or `onOptionsCallback`.
+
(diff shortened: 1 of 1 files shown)


richardhj:
Same should be added to the hooks section IMHO

docs/dev/reference/dca/callbacks.md
```diff
@@ -431,4 +431,50 @@ a button for an import "wizard".
...
+
+In case you don't use autowiring, you need to tag the corresponding service of your callback
+class:
+
+```yml
+// config/services.yml
+
+App\EventListener\DataContainer\MyTableListener:
+  tags:
+    - { name: contao.callback, table: tl_my_table, target: fields.xxMyField.options [, method: onOptionsCallback] }
```
m-vo, line 473: Do you mean auto configuration? Still, you could use the service annotation by tagging the service with `terminal42_service_annotation`.

I think it should be something like this: 
* You can define the callback directly as a tag of your service: ...
* Alternatively you can use the callback service annotation to do so: [...] 
Note: If autowiring is disabled you need to add a tag `['terminal42_service_annotation']` to your service instead of the `ServiceAnnotationInterface` marker interface in order to use the callback annotation.

Toflar: Yes, autoconfiguration not autowiring :)
And I think we should add a hint that it was only introduced in version 4.8 - I think? :) Not sure.

richardhj: I'll clarify that there are three approaches:
1. tagged service, no annotation
2. tagged service, annotation, and terminal42_service_annotation tag
3. annotation and autoconfiguration

Correct?

> And I think we should add a hint that it was only introduced in version 4.8 

I'm aware of this but information about supported versions is lacking everywhere in the documentation. I guess we assume developers to use the latest version @contao/documenters?

Toflar: Hm? No it's actually full of `{{< version "4.5" >}}` and the like hints :)

m-vo: @richardhj I think 1) should be the default - 2) & 3) are basically the same thing but with sevice annotations as a helper. 

@fritzmg I'm not sure how to deal with this in the docs in general. Maybe we would need a table/swicher widget for some things like symfony has for their various ways of writing config (yaml|xml|php)?

richardhj: When I looked at the docs for front end modules / content elements and hooks, AFAIR there was no note about the version requirements

richardhj: > @richardhj I think 1) should be the default - 2) & 3) are basically the same thing but with sevice annotations as a helper.

We should add a general section about autoconfiguration then in the docs?

fritzmg:
> Same should be added to the hooks section IMHO

Hooks already cover all three cases. Only the Service Annotation is missing for Callbacks.

docs/dev/reference/dca/callbacks.md
fritzmg: > When I looked at the docs for front end modules / content elements and hooks, AFAIR there was no note about the version requirements

There definitely is for Hooks. See https://docs.contao.org/dev/framework/hooks/#using-annotations for example.

fritzmg:
How to register callbacks is described here: https://docs.contao.org/dev/framework/dca/

richardhj:
Closing then :)

fritzmg:
You could provide a PR for the Framework article on DCAs ;). The annotation way is still missing there.

PR #134, opened 2019-10-31
Update system-requirements.de.md

Ernestopheles:
MySql Versionen ab 8.014 beinhalten neue reservierte Wörter, welches dazu führt, dass Contao nicht vollständig installiert werden kann.

### DIFF
--- a/docs/manual/installation/system-requirements.de.md
+++ b/docs/manual/installation/system-requirements.de.md
@@ -18,6 +18,7 @@ empfohlen, diese immer zu verwenden.
 
 - **PHP:** Version 7.3+ (neueste Patch-Version)
 - **MySQL:** Version 5.7+ oder gleichwertiger **MariaDB** Server
+
 {{% notice warning %}}
 Derzeit besteht eine Inkompatibilität mit MySQL Version **8.0.14+**.
 {{% /notice %}}
--- a/docs/manual/installation/system-requirements.de.md
+++ b/docs/manual/installation/system-requirements.de.md
@@ -17,7 +17,10 @@ gepflegten Versionen von Contao sind mit den neuesten PHP- und MySQL-Versionen k
 empfohlen, diese immer zu verwenden.
 
 - **PHP:** Version 7.3+ (neueste Patch-Version)
-- **MySQL:** Version 5.7+ aber vor 8.014 oder gleichwertiger **MariaDB** Server
+- **MySQL:** Version 5.7+ oder gleichwertiger **MariaDB** Server
+{{% notice warning %}}
+Derzeit besteht eine Inkompatibilität mit MySQL Version **8.0.14+**.
+{{% /notice %}}
 
 
 #### PHP-Erweiterungen
--- a/docs/manual/installation/system-requirements.de.md
+++ b/docs/manual/installation/system-requirements.de.md
@@ -20,7 +20,7 @@ empfohlen, diese immer zu verwenden.
 - **MySQL:** Version 5.7+ oder gleichwertiger **MariaDB** Server
 
 {{% notice warning %}}
-Derzeit besteht eine Inkompatibilität mit MySQL Version **8.0.14+**.
+Derzeit besteht eine Inkompatibilität mit MySQL Version **8.0.17+**.
 {{% /notice %}}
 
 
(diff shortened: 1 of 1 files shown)


fritzmg:
This would be a short lived entry.

Ernestopheles:
Hopefully... But necessary (I would have been glad to read this before
starting my project)

Am 31.10.19 um 11:40 schrieb Fritz Michael Gschwantner:

fritzmg:
> I would have been glad to read this before starting my project

The documentation cannot foresee the future and thus there can always be problems, that aren't documented. It is something you have to deal with in this field of work.

fritzmg (changes requested):
If we really want to document incompatible bugfix versions of MySQL and MariaDB, we should do this in a more detailed manner. Though then again, once a new Contao or `doctrine/dbal` version comes out, all this is pointless anyway.

Ernestopheles:
I will watch the 2.10.0 Milestone, when released I will change the documentation.

docs/manual/installation/system-requirements.de.md
```diff
@@ -17,7 +17,7 @@ gepflegten Versionen von Contao sind mit den neuesten PHP- und MySQL-Versionen k
 empfohlen, diese immer zu verwenden.
 
 - **PHP:** Version 7.3+ (neueste Patch-Version)
-- **MySQL:** Version 5.7+ oder gleichwertiger **MariaDB** Server
+- **MySQL:** Version 5.7+ aber vor 8.014 oder gleichwertiger **MariaDB** Server
```
fritzmg: I would instead insert a notice warning of the current incompatibility:

```markdown
{{% notice warning %}}
Derzeit besteht eine Inkompatibilität mit MySQL Version **8.0.17** und MariaDB Version **XYZ**.
{{% /notice %}}
```
Please research the exact MariaDB versions as well, where the new reserved keywords were introduced.

fritzmg (changes requested):

Ernestopheles:
The list of reserved words at Maria DB does not contain any of the words in question.

docs/manual/installation/system-requirements.de.md
```diff
@@ -18,6 +18,9 @@ empfohlen, diese immer zu verwenden.
 
 - **PHP:** Version 7.3+ (neueste Patch-Version)
 - **MySQL:** Version 5.7+ oder gleichwertiger **MariaDB** Server
```
fritzmg, line 20: Please add a line break after the list.

docs/manual/installation/system-requirements.de.md
```diff
@@ -18,6 +18,9 @@ empfohlen, diese immer zu verwenden.
...
+{{% notice warning %}}
+Derzeit besteht eine Inkompatibilität mit MySQL Version **8.0.14+**.
```
fritzmg: `MEMBER` was added in `8.0.17`, not `8.0.14`. See https://dev.mysql.com/doc/refman/8.0/en/keywords.html

fritzmg (changes requested):

docs/manual/installation/system-requirements.de.md
Ernestopheles: There are 3 keywords in question, see https://github.com/doctrine/dbal/pull/3674#issue-315887966

fritzmg: Only the `MEMBER` keyword is currently relevant within Contao - unless an extension uses `LATERAL` somewhere without quotes as well.

fritzmg:
I am still unsure whether we should add this to the documentation. We will not monitor _every_ MySQL and MariaDB release for certain incompatibilities that could suddenly arise - or vice versa for any incompatibilities that were reverted or otherwise fixed.

Ernestopheles:
Maybe a good idea is to also announce in the documentation, which
versions of mysql and maria have been succesfully tested with contao and
therefore are recommended for use?

Am 31.10.19 um 13:05 schrieb Fritz Michael Gschwantner:

fritzmg:
That is not feasibly either, as there are too many different versions.

Magento keeps a list of MySQL and MariaDB minor versions that have been tested for example. But this would not have helped you in this case, since the breaking change was introduced by MySQL in a _bugfix_ version.

Ernestopheles:
I know, in general software users have to search through forums to find
out about bugs they experience, and I do not like it. So contao is here
not worse than probably most others. Software-engineering in general is
far from being perfect. Users will happily notice when this changes one day.


Am 31.10.19 um 13:25 schrieb Fritz Michael Gschwantner:

docs/manual/installation/system-requirements.de.md
Ernestopheles: Thanks, done

netzarbeiter:
Thank you. I'm closing this PR, because we don't want to add short lived entry to the docs.

PR #138, opened 2019-10-31
Added initial version of our BC promise

Toflar:
This is quite an important one so I've requested a few people for review.
Please feel free to notify other Contao devs if you think they would like to extend/modify this document.

### DIFF
--- a/docs/dev/internals/bc-promise.md
+++ b/docs/dev/internals/bc-promise.md
@@ -8,16 +8,16 @@ We try our very best to ensure that Contao updates are as smooth as they can be.
 
 Being a project that is built on top of the Symfony full-stack framework, we generally make the same
 [Backwards Compatibility Promise][SF_BC_Promise] as Symfony itself.
-We also follow the principles of [Semantic Versioning][Semver], which means that breaking changes must only be expected
-when switching to a new major version.
+We also follow the principles of [Semantic Versioning][Semver], which means that breaking changes must only be
+expected when switching to a new major version.
 
 However, in contrast to Symfony, Contao is not "just" a framework. We do not only build the tools for other
 developers but we also build tools ourselves, which are then shipped with the core distribution of Contao.
 Therefore, our BC promise deviates from the Symfony BC promise in some regards:
 
-* Any method (constructors included) that is marked as `@internal` is not covered. In most cases this concerns constructors
-  of services Contao provides. If you want to change the behaviour of a service, don't replace it by instantiating your
-  own instance of a class but instead decorate the original service. Also see the tip about "Composition over Inheritance".
+* Our BC promise does not cover classes and methods marked as `@internal`. In most cases this concerns constructors
+  of services Contao provides. If you want to change the behaviour of a service, do not replace it with your own
+  instance of the class but instead decorate the original service (see the tip about "Composition over Inheritance").
 
 * Our BC promise does not cover templates. Templates are subject to change very often and you have to compare
   them with every update of Contao. Generally, we try to only ever apply template changes in major and minor
@@ -31,9 +31,9 @@ Therefore, our BC promise deviates from the Symfony BC promise in some regards:
   integrating Contao into the Symfony application. This includes:
   
   * Commands
-  * DataCollectors
-  * Dependency Injection Compiler Passes
-  * Event Listeners
+  * Data collectors
+  * Dependency injection compiler passes
+  * Event listeners
 
 * Our BC promise also does not cover the `ContaoManager/Plugin` class, which is required to integrate a bundle into
   the [Contao Managed Edition][Contao_ME].
--- a/docs/dev/internals/bc-promise.md
+++ b/docs/dev/internals/bc-promise.md
@@ -15,6 +15,10 @@ However, in contrast to Symfony, Contao is not "just" a framework. We do not onl
 developers but we also build tools ourselves, which are then shipped with the core distribution of Contao.
 Therefore, our BC promise deviates from the Symfony BC promise in some regards:
 
+* Any method (constructors included) that is marked as `@internal` is not covered. In most cases this concerns constructors
+  of services Contao provides. If you want to change the behaviour of a service, don't replace it by instantiating your
+  own instance of a class but instead decorate the original service. Also see the tip about "Composition over Inheritance".
+
 * Our BC promise does not cover templates. Templates are subject to change very often and you have to compare
   them with every update of Contao. Generally, we try to only ever apply template changes in major and minor
   versions but if a bugfix requires us to change a template, then we might do that as well.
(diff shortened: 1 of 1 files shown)


docs/dev/internals/bc-promise.md
```diff
@@ -0,0 +1,45 @@
...
+  application might be subject to change. So expect breaking changes for the following aspects of Contao:
+  
+  * Event Listeners
+  * Dependency Injection Compiler Passes
+  * Commands
+  
+
+The general rule of thumb is that we try to break as little as possible and as much as required. To make sure you're
+affected as little as possible, you may also benefit from best practices in software engineering such as preferring
+Composition over Inheritance][Composition_over_Inheritance].
```
bytehead: Missing `[`

Toflar: Fixed

m-vo:
Let's talk constructor parameters again.

Personally I think the exact dependencies I'm injecting into a class (that do not leak out via class methods) should be considered implementation detail: If I want to change an algorithm for instance, I'd like to be able to exchange `LibraryAGenerator` with `LibraryBProvider`, if I want to improve 'testability', I'd like to insert new dependencies - e.g. a `FilesystemAbstraction`.

So, when does this hurt anyone? Basically whenever there are calls to my class's constructor with a now changed (and incompatible = no respective optional parameters) signature due to...
1) ... someone creating an instance of my class.
2) ... someone calling `parent::__construct()`.

In a DI architecture where classes (services) are instantiated by the container 1) can be neglected. 2) is only the case when inheriting and there are only a few reasons someone should ever do this:

1. extending an abstract base class (`Contao\CoreBundle\Controller\ContentElement\AbstractContentElementController`)
2. extending a non-abstract class that is meant to be a building block (`Symfony\Component\Console\Command\Command`)
3. creating a fixture for testing (`Contao\CoreBundle\Fixtures\Exception\DerivedPageNotFoundException`)
4. overwriting features of a class (e.g. decorating a service & modifying bits)

Points 1) - 3) are legitimate use cases. 4), however, clearly is a misuse to me (the class is maybe simply not `final` due to point 4).

So how to work around this? Let's make an example: If I want decorate Doctrine's default `AnnotationReader` instead of  extending and overwriting it I could also implement the `Reader` interface myself. And If I need to reuse code can inject the existing `AnnotationReader` and simply delegate calls. This composition pattern works whenever there are appropriate interfaces around.

Long story short, because all of you know where this is going anyway: :smile:  I would vote for **excluding constructors from BC promise** and make **some well-thought exceptions** to this rule, namely to cover the above building blocks. Note that this is only true for the 'new' code, not the legacy classes inside `Resources/contao`.

(And imho those exceptions should be visible in code - e.g. with appropriate annotations. )

docs/dev/internals/bc-promise.md
```diff
@@ -0,0 +1,45 @@
+---
+title: "BC Promise"
+description: "Our backwards compatibility promise."
+---
+
+Everyone involved in the development of Contao knows about the importance of stable software and thus stable APIs.
+We try our very best to make sure Contao updates are as smooth as they can be.
+
+Being a project that sits on top of the Symfony full-stack framework we generally [follow the same rules as
```
xchs:
```suggestion
Being a project that sits on top of the Symfony full-stack framework, we generally [follow the same rules as
```

docs/dev/internals/bc-promise.md
```diff
@@ -0,0 +1,45 @@
...
+Symfony itself][SF_BC_Promise].
+We also follow the principles of [Semantic Versioning][Semver] which means that breaking changes must only be expected
```
xchs:
```suggestion
We also follow the principles of [Semantic Versioning][Semver], which means that breaking changes must only be expected
```

docs/dev/internals/bc-promise.md
```diff
@@ -0,0 +1,45 @@
...
+when switching to a new major version.
+
+However, in contrast to Symfony itself, Contao is not "just" a framework. We do not only build the tools for other
+developers but we also build tools ourselves which are then shipped with the core distribution of Contao.
```
xchs:
```suggestion
developers but we also build tools ourselves, which are then shipped with the core distribution of Contao.
```

qzminski (dismissed):

xchs (dismissed):
👍

leofeyer:
I like the initial version and especially the following sentence:

> everything that is about integrating Contao into a Symfony application might be subject to change

What we could discuss:

1. Constructor arguments (see @m-vo's arguments above)
2. Composer/ScriptHandler (integrating Contao into a Symfony application)
3. ContaoManager/Plugin (integrating Contao into the Managed Edition)

Toflar:
> Constructor arguments (see @m-vo's arguments above)

I don't think I will ever agree here.



> Composer/ScriptHandler (integrating Contao into a Symfony application)

I don't think we need to mention this here. We won't be able to change that in any minor version so that must be covered by BC.



> ContaoManager/Plugin (integrating Contao into the Managed Edition)

Added.

m-vo:
> don't think I will ever agree here.

Do you mind sharing why?

Toflar:
Symfony protects all public methods that obviously includes constructors. Which makes sense because anybody can register their own services and use the classes the way they want.
And that's also the way it should be.
We cannot just adjust the constructor of e.g. a `PictureGenerator`. I might use my own instance of it with a different instance of `ResizerInterface`.

richardhj (dismissed):

fritzmg (dismissed):

ausi:
I’m not sure about the constructor arguments. For services it might make sense to exclude them from the BC promise. Should we set this to “up for discussion”?

leofeyer:
There is no "up for discussion" label in this repo, but we should definitely discuss this. I have set up a reminder for the next call.

aschempp:
I generally like the statement(s). Some thoughts:
 - we might mentions that the template _markup_ can change, but the template variables should not (the should only be extended).
 - There is at least one compiler pass that is explicitly designed to be re-used by extension developers: https://github.com/contao/contao/blob/master/core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php#L29

leofeyer (approved):

leofeyer (changes requested):
As discussed in Mumble on November 28th, we want to add a paragraph about constructors marked as `@internal` before we merge.

leofeyer (dismissed):

richardhj (dismissed):

ausi (dismissed):

docs/dev/internals/bc-promise.md
```diff
@@ -0,0 +1,56 @@
...
+We also follow the principles of [Semantic Versioning][Semver], which means that breaking changes must only be
+expected when switching to a new major version.
+
+However, in contrast to Symfony, Contao is not "just" a framework. We do not only build the tools for other
+developers but we also build tools ourselves, which are then shipped with the core distribution of Contao.
+Therefore, our BC promise deviates from the Symfony BC promise in some regards:
+
+* Our BC promise does not cover classes and methods marked as `@internal`. In most cases this concerns constructors
+  of services Contao provides. If you want to change the behaviour of a service, do not replace it with your own
+  instance of the class but instead decorate the original service (see the tip about "Composition over Inheritance").
```
aschempp, line 20: Can we somehow link to https://symfony.com/doc/current/service_container/service_decoration.html here?

docs/dev/internals/bc-promise.md
```diff
@@ -0,0 +1,56 @@
...
+* Because Contao is a Symfony bundle like every other bundle, our BC promise does not cover anything that is about 
+  integrating Contao into the Symfony application. This includes:
+  
+  * Commands
+  * Data collectors
+  * Dependency injection compiler passes
+  * Event listeners
+
+* Our BC promise also does not cover the `ContaoManager/Plugin` class, which is required to integrate a bundle into
+  the [Contao Managed Edition][Contao_ME].
```
aschempp, line 39: I'm not sure this whole chapter is necessary if we simply mark them as `@internal`? All of the mentioned class types (except the Plugin) are also services, so they are already covered by _you should decorate a service_?

leofeyer (approved):

PR #146, opened 2019-11-05
Fixed a few typos

Toflar:


fritzmg:
I always considered "Insert Tag" to be a Contao specific construct and thus handled it as a name. But same goes for "hook(s)" and we don't (always?) write it as a name as well.

fritzmg (approved):

Toflar:
Same for me but it was never a name for @leofeyer :)
I would‘ve even gone for `InsertTag` for all languages but that was denied multiple times. It‘s not super relevant anyway.

PR #150, opened 2019-11-06
Fix broken link

fiedsch:
The Link is broken and this is what I think it should be.


fritzmg:
I am not sure why that's even really in there. I have created a different proposal.

Toflar (approved):

fritzmg:
Thank you @fiedsch :)

PR #151, opened 2019-11-06
Use cronjob to label pull requests

Toflar:
See https://github.com/actions/labeler/issues/12 for details.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,7 +1,8 @@
 name: Label the Pull Requests automatically
 
 on:
-    - pull_request
+    schedule:
+        - cron: '0 * * * *'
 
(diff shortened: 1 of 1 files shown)


fritzmg:
Shouldn't the interval be shorter? Chances are that I will always be faster 😁

Toflar:
I didn't want to hammer GitHub but we can also shorten the interval. Suggestions?

fritzmg:
May be 15 min? But yeah... auto-labeling isn't that important, so we might as well leave it at 1 hour, so as to not hammer GitHub and waste resources.

fritzmg (approved):

Toflar:
I would leave it like that for now and see how it goes.

PR #155, opened 2019-11-06
Various typos and improvements for publishing bundles

Toflar:


docs/dev/guides/publishing-bundles.md
```diff
@@ -59,19 +59,19 @@ If you do **not** add a _Contao Manager Plugin_ to your extension, it will still
...
+   to the Contao Managed Edition, and won't allow its installation with
+   a push of a button.
 
 
 ## Publishing to the Contao ecosystem
 
-If you followed so far, your extension is now technically compatible with Contao, 
-but you also want Contao users to discover it. In most cases, this means having the package
-available in the Contao Manager search. A public version of this search index
+If you've been following along so far, your extension is now technically compatible with Contao
```
fritzmg: I think we should avoid contractions like that in the documentation, shouldn't we?

fritzmg (changes requested):

docs/dev/guides/publishing-bundles.md
Toflar: Right

Toflar: I've updated the wording.

fritzmg (approved):

PR #191, opened 2019-11-28
Add guide on how to use a custom compiler pass in the Managed Edition

m-vo:
TLDR; 

The [Symfony documentation for adding compiler passes](https://symfony.com/doc/4.4/service_container/compiler_passes.html) doesn't help in a bundle less managed edition application without access to the Kernel. But the Manager Plugin can be used instead.

### DIFF
--- a/docs/dev/guides/me-compiler-pass.md
+++ b/docs/dev/guides/me-compiler-pass.md
@@ -22,8 +22,8 @@ container and thus also add compiler passes there.
 ### Contao Manager Plugin to the rescue
 
 Make sure your plugin implements the `ConfigPluginInterface` and implement the
-`registerContainerConfiguration` method. Now the provided loader allows
-accessing the container and adding your compiler pass:
+`registerContainerConfiguration` method (see [Plugin documentation](/framework/manager-plugin#the-configplugininterface)).
+Now the provided loader allows accessing the container and adding your compiler pass:
 
 ```php
 // src/ContaoManager/Plugin.php
--- a/docs/dev/guides/me-compiler-pass.md
+++ b/docs/dev/guides/me-compiler-pass.md
@@ -1,0 +1,46 @@
+---
+title: "Add Custom Compiler Passes (ME)"
+description: "How to add a custom compiler pass in the Contao Managed Edition."
+aliases:
+  - /guides/me-compiler-pass.md
+---
+
+
+## Add Custom Compiler Passes
+
+In a regular Symfony application or bundle a compiler pass can be added by calling
+`$container->addCompilerPass()` inside the `build()` function of the application's
+`Kernel` or the bundle's `Bundle` class (see the [Symfony documentation](https://symfony.com/doc/4.4/service_container/compiler_passes.html)
+for details).
+ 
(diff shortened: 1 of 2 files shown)


Toflar (commented):
That's not only true for all the compiler passes but basically whenever you want to do something with the container :) But I like we have a guide. Should maybe also link to the plugin page?

docs/dev/guides/me-compiler-pass.md
```diff
@@ -0,0 +1,46 @@
...
+In a typical **Contao Managed Edition** setup, we cannot alter the Kernel nor does the
+application - following current best practices to be *bundle-less* - contain a
+`Bundle` class. However, we can also utilize the Contao Manager Plugin to alter the
+container and thus also add compiler passes there.
+
+
+### Contao Manager Plugin to the rescue
+
+Make sure your plugin implements the `ConfigPluginInterface` and implement the
+`registerContainerConfiguration` method. Now the provided loader allows
```
fritzmg: Please add a link to the plugin's documentation here :)

fritzmg (changes requested):

m-vo:
> That's not only true for all the compiler passes but basically whenever you want to do something with the container :) 

I thought that as well when writing. Should I rename the file to something like "altering the container at compile time in the ME"?

docs/dev/guides/me-compiler-pass.md
```diff
@@ -0,0 +1,46 @@
+---
+title: "Add Custom Compiler Passes (ME)"
+description: "How to add a custom compiler pass in the Contao Managed Edition."
+aliases:
+  - /guides/me-compiler-pass.md
```
fritzmg: Alias should be
```
/guides/me-compiler-pass/
```
Also I think the filename should be `add-compiler-pass`. Yes, it is about the managed addition, but I would still like that filename/slug better ;)

fritzmg (changes requested):

docs/dev/guides/me-compiler-pass.md
m-vo: I changed the name/content of the guide to be more general - could you check again if you like the naming and if I did the right thing with the alias?  :slightly_smiling_face:

docs/dev/guides/modify-container-at-compile-time.md
```diff
@@ -0,0 +1,49 @@
+---
+title: "Modifying the Container at Compile Time (ME)"
```
fritzmg:
```suggestion
title: "Modifying the Container at Compile Time"
```

😁

docs/dev/guides/modify-container-at-compile-time.md
```diff
@@ -0,0 +1,49 @@
...
+description: "How to make changes to the container at compile time in the Contao Managed Edition."
+aliases:
+  - /guides/modify-container-at-compile-time/
+---
+
+
+## Modifying the container at compile time
```
fritzmg:
```suggestion
## Modifying the Container at Compile Time
```

fritzmg (changes requested):

fritzmg (approved):

fritzmg:
thank you @m-vo

PR #195, opened 2019-12-03
Add annotation reference to fragment controller documentation

fritzmg:
Annotations were only mentioned as a side note so far. This adds a more complete reference to those articles.

### DIFF
--- a/docs/dev/framework/caching.md
+++ b/docs/dev/framework/caching.md
@@ -1,6 +1,8 @@
 ---
 title: "Caching"
 description: "HTTP caching in Contao"
+alias:
+  - /framework/caching/
 ---
 
 
@@ -215,8 +217,27 @@ Moreover, you don't have to register to all the different callbacks such as `ons
 You can register to the [`oninvalidate_cache_tags` callback][5] and add your own tags.
 
 
+## Fragments and Edge Side Includes
+
+In Contao, content elements and front end modules can be implemented as so called
+_fragment controllers_. These fragments are then rendered with their defined renderer 
+and merged into the main content. Each fragment can also provide its own response
+and thus define whether it can be cached or not. If a fragment returns a response
+with a `Cache-Control: private` header for example, then the page on which the fragment
+is visible cannot be cached. On the other hand, if the fragment can be cached, it
+can provide its own cache tags, as mentioned previously.
+
+Contao brings its own `forward` fragment renderer, which provides the fragment with 
+a full clone of the request (contrary to Symfony's default `inline` renderer). The 
+renderer can also be set to `esi`. In that case, if Symfony detects that it is talking 
+to  a gateway cache that supports ESI (like Symfony's built in reverse proxy, that 
+Contao uses), it generates an ESI include tag. See also [Symfony's documentation][esi] 
+on _Edge Side Includes_.
+
+
 [1]: https://github.com/Toflar/psr6-symfony-http-cache-store
 [2]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching
 [3]: https://foshttpcachebundle.readthedocs.io/en/latest/
 [4]: https://symfony.com/doc/current/components/http_kernel.html
-[5]: ../../reference/dca/callbacks/#config-oninvalidate-cache-tags
+[5]: /reference/dca/callbacks/#config-oninvalidate-cache-tags
+[esi]: https://symfony.com/doc/current/http_cache/esi.html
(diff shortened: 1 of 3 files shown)


Toflar (dismissed):
Nice one!

docs/dev/framework/content-elements.md
```diff
@@ -143,6 +143,47 @@ $GLOBALS['TL_LANG']['CTE']['my_content_element'] = [
...
+{{< version "4.8" >}}
+
+Instead of tagging the content element controller service via the service configuration,
+the service tag can also be configured through annotations, as already used in the 
+code example above. The annotation can be used on the class of the content element 
+or on the method that will deliver the response.
+
+The following example sets the type of the content element to `my_example`, puts 
+it in the `media` category, sets the template name to `ce_some_example` and defines
+the renderer to be `inline` (which is the default):
```
aschempp: That's not exactly true, Contao ships with its own renderer (`forward`). I would either mention `esi` or none at all, as most users should not use that anyway 🙈

docs/dev/framework/content-elements.md
```diff
@@ -143,6 +143,47 @@ $GLOBALS['TL_LANG']['CTE']['my_content_element'] = [
 ```
 
 
+## Annotation
+
+{{< version "4.8" >}}
+
+Instead of tagging the content element controller service via the service configuration,
+the service tag can also be configured through annotations, as already used in the 
+code example above. The annotation can be used on the class of the content element
```
aschempp: > can be used on the class

if that class is invokable (has `__invoke` method), or basically if it extends the abstract controller

aschempp (changes requested):

docs/dev/framework/content-elements.md
fritzmg: Hmmm... I am confused now, in the core `inline` is used as well as `forward`.

aschempp: Not sure what you're referring to? The default renderer for Contao fragments is `forward`: https://github.com/contao/contao/blob/master/core-bundle/src/DependencyInjection/Compiler/RegisterFragmentsPass.php#L106

fritzmg: In the FragmentHandler the default is `inline`:

https://github.com/contao/contao/blob/fd61bc4e34608f38112f97dfa1723337362ba229/core-bundle/src/Fragment/FragmentHandler.php#L66

But yeah, when registering a fragment controller, it would default to `forward` obviously. However, that means that the possible values are `forward`, `inline` and `esi`, correct?

aschempp: there's more than these three (like `hinclude` or any custom one). But these three are the only ones that make sense. Or maybe `inline` never makes sense now that we have `forward` 🤷‍♂

fritzmg:
To make the `renderer` options more clear, I have expanded this PR with the following:

* All three renderer options mentioned.
* Added a new sections in the Caching documentation about fragments & ESI.

I also added a brief note mentioning how to implement content elements and front end modules in previous Contao versions.

@Toflar @aschempp pls review :)

docs/dev/framework/caching.md
```diff
@@ -215,8 +217,27 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+In Contao, content elements and front end modules can be implemented as so called
+_fragment controllers_. These fragments are then rendered with their defined renderer 
+and merged into the main content. Each fragment can also provide its own response
+and thus define whether it can be cached or not. If a fragment returns a response
+with a `Cache-Control: private` header for example, then the page on which the fragment
+is visible cannot be cached. On the other hand, if the fragment can be cached, it
+can provide its own cache tags, as mentioned previously.
+
+Contao brings its own `forward` fragment renderer, which provides the fragment with 
+a full clone of the request (contrary to Symfony's default `inline` renderer). The
```
Toflar: The info missing here is that it will also provide the `POST` data, in contrary to the `inline` renderer.

Toflar (commented):
At some point, I would also like to make some matrix that should help the developers understand when to choose ESI :) It's a really difficult topic to understand. The fact that it's a fragment is easy to grasp but the implications aren't.

fritzmg:
> At some point, I would also like to make some matrix that should help the developers understand when to choose ESI :) It's a really difficult topic to understand. The fact that it's a fragment is easy to grasp but the implications aren't.

Agreed. The documentation is a living thing anyway :)

docs/dev/framework/caching.md
fritzmg: I have updated the wording.

aschempp (approved):

PR #203, opened 2019-12-19
Improved introduction to the Manager Plugin

Toflar:
I noticed that the introduction of the Manager Plugin was only talking about "third party bundles". The application specific one is only mentioned later.

### DIFF
--- a/docs/dev/framework/manager-plugin.md
+++ b/docs/dev/framework/manager-plugin.md
@@ -1,6 +1,6 @@
 ---
 title: Manager Plugin
-description: "A class that gives third party bundles the opportunity to configure a Contao Managed Edition."
+description: "A class that allows third party bundles and application developers to configure a Contao Managed Edition."
 aliases:
     - /framework/manager-plugin/
     - /framework/managed-edition/manager-plugin/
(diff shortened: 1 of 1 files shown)


fritzmg:
I'd like to normalize how things should be emphasized:

* Single back ticks should only be used on the following:
  * Code snippets
  * Namespaces
  * Class and function names
  * Commands
  * Log and error messages
* Emphasizing a word should be done with `_` or `__`, as this also creates the semantically correct `<em>` or `<strong>` tag.

Things like _Contao Managed Edition_, _Manager Plugin_ etc. should be emphasized with `_` or `__`.

fritzmg (changes requested):
I'd like to normalize how things should be emphasized:

* Single back ticks should only be used on the following:
  * Code snippets
  * Namespaces
  * Class and function names
  * Commands
  * Log and error messages
* Emphasizing a word should be done with `_` or `__`, as this also creates the semantically correct `<em>` or `<strong>` tag.

Things like _Contao Managed Edition_, _Manager Plugin_ etc. should be emphasized with `_` or `__`.

Toflar:
I see but that makes no sense in this PR because the rest of the document would need to be normalized as well and that's not related to the introduction-rewrite :)

fritzmg:
Fair enough ;)

fritzmg (approved):

PR #206, opened 2020-01-02
Make hook URLs camel case

fritzmg:
This PR changes the Hugo config to allow URLs with upper case letters, to retain the camel case of hooks in their URLs. This also allows then to use the following rewrite rule within the `.htaccess` of the docs:

```
RewriteRule ^books/api/extensions/hooks/([a-zA-Z]*).html$    /dev/reference/hooks/$1/ [R=301,L]
```

This PR also adds appropriate aliases to each hook reference article, so that a redirect from the old lower case URL is made.

### DIFF
--- a/docs/dev/reference/hooks/activateAccount.md
+++ b/docs/dev/reference/hooks/activateAccount.md
@@ -2,6 +2,9 @@
 title: "activateAccount"
 description: "activateAccount hook"
 tags: ["hook-module", "hook-member"]
+aliases:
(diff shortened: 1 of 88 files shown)


netzarbeiter (approved):
Thank you

fritzmg:
Redirects are active too now. e.g.: https://docs.contao.org/books/api/extensions/hooks/insertTagFlags.html

PR #211, opened 2020-01-07
Update front-end-modules.md

hendrikbengs:
Changed {type_legend} to {title_legend} because there is no {type_legend} defined in tl_module.php; also add the dca-field name as default


fritzmg (approved):

fritzmg:
Thank you @HendrikBengs

PR #212, opened 2020-01-07
Documented the search indexer abstraction introduced in Contao 4.9

Toflar:


docs/dev/framework/search-indexing.md
```diff
@@ -0,0 +1,150 @@
...
+
+This is why you can configure the behaviour of the `SearchIndexListener` as follows:
+
+```yaml
+# config/config.yml
+contao:
+    search:
+        listener:
+            index: true # Configure whether you want to update the index entry on every request
+            delete: false # Configure whether you want to delete the obsolete entry on every request
```
leofeyer:
```suggestion
            delete: false # Configure whether you want to delete an entry if a request is not successful
```

leofeyer (dismissed):

leofeyer (approved):

fritzmg (approved):

PR #215, opened 2020-01-08
Fixed storeFormData example to not use legacy stuff anymore

### LINKED ISSUE #214: update source code - deprecated methodes
Change `FE_USER_LOGGED_IN` and database call in  https://docs.contao.org/dev/reference/hooks/storeFormData/#example

See https://contao.slack.com/archives/CNZBHDCP4/p1578417426012200

Toflar:
Fixes https://github.com/contao/docs/issues/214

### DIFF
--- a/docs/dev/reference/hooks/storeFormData.md
+++ b/docs/dev/reference/hooks/storeFormData.md
@@ -38,10 +38,29 @@ namespace App\EventListener;
 
 use Contao\CoreBundle\ServiceAnnotation\Hook;
 use Contao\Form;
+use Contao\FrontendUser;
+use Doctrine\DBAL\Connection;
+use Symfony\Component\Security\Core\Security;
 use Terminal42\ServiceAnnotationBundle\ServiceAnnotationInterface;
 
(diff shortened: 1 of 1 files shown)


zonky2:
THX

fritzmg:
`\Contao\Database` isn't actually deprecated and contains useful features, which would make the example much simpler. Instead of having to write your own `columnExistsInTable` function from scratch, you can simply use `fieldExists` of `\Contao\Database`.

Toflar:
You shouldn't be using that class anymore anyway. Even if it's useful, just don't. I haven't used it anymore for over 4 years now. We may provide some util classes in the core someday but I just don't want the docs to contain old stuff if possible.

fritzmg (approved):

fritzmg:
Thank you @Toflar

zonky2:
thx!

Toflar:
@fritzmg maybe we could use `$columns = $schemaManager->listTableColumns($form->targetTable);`. Then there's no need for my "complex" separate method.

PR #219, opened 2020-01-11
Fix tree view params

Tastaturberuf:
see https://github.com/contao/contao/blob/master/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3995
and https://github.com/contao/contao/blob/master/core-bundle/src/Resources/contao/drivers/DC_Table.php#L3999


fritzmg (approved):

fritzmg:
Thank you @Tastaturberuf

PR #224, opened 2020-01-15
Added new documentation for back end menu item (Contao 4.9 or higher)

fabcocco:
Creating Backend Menu Item in Contao 4.9

### DIFF
--- a/docs/dev/guides/backend-routes.md
+++ b/docs/dev/guides/backend-routes.md
@@ -112,7 +112,7 @@ Since the back end menu can be extended with an `EventListener` we can easily
 create one that listens for the menu event to be dispatched.
 
 {{% notice warning %}}
-The following functions works in Contao **>4.9** for the Version **4.9** use the BackendMenuListener.php from above.
+The following example only works in Contao **<4.9**. For >=4.9 see the example below. 
 {{% /notice %}}
 
 ```php
--- a/docs/dev/guides/backend-routes.md
+++ b/docs/dev/guides/backend-routes.md
@@ -5,11 +5,6 @@ aliases:
   - /guides/backend-routes
 ---
 
-{{% notice note %}}
-This covers the documentation on how to create routes in Contao, the old Contao **3.5** way. The Recomondation for Contao **4.9 LTS** and up is a different.
-Be aware of two different ways: local Extension & Bundles.
-{{% /notice %}}
-
 ## Adding custom back end routes
 
 You can use the Contao back end to display content generated in your own custom Controllers.
@@ -17,7 +12,6 @@ This way you can develop custom extensions without the need to use DCA configura
 The following example can be changed according to your own setup. For example you're
 not obliged to use the annotation configuration for your routes you could use
 XML or YAML interchangeably.
-
 
 ### Create your Controller and Template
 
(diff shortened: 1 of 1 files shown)


fritzmg:
Thank you for your contribution :). Your pull request contains formatting changes as well as new content. It would be better make two pull requests: one for improving the formatting and one for the additional new content. Also your Pull Request title should be changed to something more useful ;)

theKalash:
We removed the "improved formatting" for now ;)

docs/dev/guides/backend-routes.md
```diff
@@ -111,6 +111,10 @@ Most of the time you probably want to add a menu entry for your back end module.
 Since the back end menu can be extended with an `EventListener` we can easily
 create one that listens for the menu event to be dispatched.
 
+{{% notice warning %}}
+The following functions works in Contao **>4.9** for the Version **4.9** use the BackendMenuListener.php from above.
```
fritzmg: Don't you mean it only works in Contao 4.9 and previous versions need to use the example above?

fritzmg (changes requested):

docs/dev/guides/backend-routes.md
theKalash: Yes!

theKalash: and above should probably be "below"  referring to the new block we added

theKalash: The original example only works for Versions up to 4.8.* (before the menu refactoring), the example we added is for 4.9 and above.

fritzmg:
@leofeyer but isn't it a BC break, that the previous example does not work any more?

theKalash:
So from my knowledge, this is was changed between 4.8 and 4.9:

With 4.8 the call to `$event->getTree();` returned the sidebar menu. 
With 4.9 it returns the top-bar menu, and thus `$contentNode = $tree->getChild('content');` returns null and calling `->addChild()` fails. 

We also noticed that now that we add our own content Node, it appears in both, the top- and the sidebar.

leofeyer:
Here is how you retrieve the sidebar menu in Contao 4.9:

```php
    public function addCustomItem(MenuEvent $event): void
    {
        $tree = $event->getTree();

        if ('mainMenu' !== $tree->getName()) {
            return;
        }

        $contentNode = $tree->getChild('content');
        // ...
```

The existing example only works accidentally, because in Contao 4.8 there is only one menu. 🤷‍♂

fritzmg:
So is there a way that works in Contao 4.4, 4.8 and 4.9? If yes, then only that way should get documented.

leofeyer:
Well, the menu name in Contao 4.8 was `root` instead of `mainMenu`, so there is no identical solution for both versions. However, Contao 4.8 will reach its EOL in 4 weeks, so we should only document how it works in Contao 4.9.

theKalash:
Well inserting 

```
        if ('mainMenu' !== $tree->getName()) {
            return;
        }
```
Should be all that is necessary to update the existing example for 4.9. So I think we can disregard this PR

Though, using the setters on the factory seems much nicer than the array structure that was used before.

fritzmg:
> Well, the menu name in Contao 4.8 was `root` instead of `mainMenu`, so there is no identical solution for both versions. However, Contao 4.8 will reach its EOL in 4 weeks, so we should only document how it works in Contao 4.9.

I see, so this was not possible in Contao 4.4 at all? Yeah, then we should simply change the current example to how it would work in 4.9.

theKalash:
We updated it to only include the condition for the mainMenu and the new method based syntax. That should fix it for 4.9 and make it slightly prettier.

fritzmg (approved):

PR #226, opened 2020-01-16
Update caching documentation for Contao 4.9 and inline fragments

aschempp:


leofeyer (dismissed):

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+
+There are two fundamentally different ways to render fragments:
+ 1. **Inline:** Inside the main page request and merged with the content before caching. In Symfony, this is handled
+    by the `inline` renderer, but Contao provides its own default `forward` renderer, which copies the original
+    request (whereas Symfony does not) to e.g. pass POST data to the subrequest.
+ 2. **Subrequest:** By generating an (internal) URL for the fragment, the fragment content is merged **after** caching
+    the page, which allows a fragment to have a different cache time than the main page. The `esi` renderer tells
+    the reverse proxy (cache) to fetch multiple URLs on a request and merge fragments before delivering to the client.
+    There's also an `hinclude` renderer, which merges fragments in the browser using JavaScript, but there's hardly
+    any use case for this anymore today.
```
Toflar: That‘s not true. Quite the opposite is true. There are many use cases for client side fragments for caching purposes (e.g. a personal shopping cart). But you don‘t need hinclude for this, you can just develop the JS/Ajax solution yourself.

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+
+### Inline Fragments
+
+{{% notice note %}}
+Before **Contao 4.9**, inline fragments cannot affect the cache time of the page response.
+{{% /notice %}}
+
+By default, fragments like frontend modules and content elements are rendered directly inside the main request.
+This means the content of an inline fragment will be cached for as long as the page is cached, per definition
+in the page setting. If a fragment returns caching information in it's response, the page's cache time and the fragment
```
Toflar:
```suggestion
in the page setting. If a fragment returns caching information in its response, the page's cache time and the fragment
```

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+
+
+### Inline Fragments
+
+{{% notice note %}}
+Before **Contao 4.9**, inline fragments cannot affect the cache time of the page response.
+{{% /notice %}}
+
+By default, fragments like frontend modules and content elements are rendered directly inside the main request.
+This means the content of an inline fragment will be cached for as long as the page is cached, per definition
```
Toflar: `as configured in the page settings` maybe?

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+
+By setting the fragment renderer to `esi`, the fragment will be cached **separately** from the main content.
+The page can be cached for 24 hours, but the frament can be cached for one week. On every request to the page,
+the reverse proxy will merge the two pieces, rebuilding each if its cache has expired.
+
+A common use case for this is a fragment that can certainly be cached longer than the current page, but is
+expensive to generate. Something like a weather preview, which requires an API request, but only updates once a day.
+Do not use ESI if your fragment is inexpensive to generate (like the `{{date::Y}}` insert tag). For inexpensive
+cases, it is most likely better to cache the whole page and **re-generate more often** than having to merge multiple
+fragments **on each request**.
```
Toflar, line 352: `on every request`. You should also mention why. The explanation of how ESI works is missing and so it‘s not understood that the contents have to be parsed on every request for newcomers.

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+
+Symfony automatically detects if it is talking to  a gateway cache that supports ESI
+(like Symfony's built in reverse proxy, that Contao uses). ESI is also supported by
```
Toflar: Only the Managed Edition.

Toflar (changes requested):

docs/dev/framework/caching.md
aschempp: that's why `hinclude` is obsolete today :wink:

docs/dev/framework/caching.md
aschempp: that should be explained in https://github.com/contao/docs/pull/226/files#diff-7c61bf1c962ca5b3891de53ec54776acR232 ?

docs/dev/framework/caching.md
aschempp: yeah… thats how the docs were before. Not sure if that detail matters?

Toflar: It does.

docs/dev/framework/caching.md
Toflar: It's not at all. Its' perfectly fine to use it. Contao just doesn't automatically include it in the layout. Which it actually could if there was an `hinclude` fragment 🤔

docs/dev/framework/caching.md
Toflar: Please still add some more context. People don't read the whole documentation but only parts.

docs/dev/framework/caching.md
aschempp: not sure what your point is. Practically, `hinclude` is irrelevant for Contao and this documentation. So there's no reason to document it.

docs/dev/framework/caching.md
aschempp: that's literally a few lines above. Should we like to "see above"? 😂

docs/dev/framework/caching.md
Toflar: But the "hardly any use case" statement is just plain wrong.

fritzmg:
@aschempp can you update this PR?

fritzmg:
@aschempp @Toflar can you look over this again?

docs/dev/framework/caching.md
aschempp: changed in 9c27ffaf49e6e014e21c9f363e399ebe84f2140f

docs/dev/framework/caching.md
aschempp: Changed in 9c27ffaf49e6e014e21c9f363e399ebe84f2140f

docs/dev/framework/caching.md
aschempp: Added in 204e48085716e72b5a31785de634f913592f381b

Toflar (commented):
I don't know what to do with this PR, really. A lot of the information in this document is technically not a 100% correct. It's better than what we have now, though.

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+_fragment controllers_. These fragments are then rendered with their defined renderer
+and merged into the main content. Each fragment returns a response and thus can
+tell whether it can be cached or not.
+
+
+There are two fundamentally different ways to render fragments:
+ 1. **Inline:** Inside the main page request and merged with the content before caching. In Symfony, this is handled
+    by the `inline` renderer, but Contao provides its own default `forward` renderer, which copies the original
+    request (whereas Symfony does not) to e.g. pass POST data to the subrequest.
+ 2. **Subrequest:** By generating an (internal) URL for the fragment, the fragment content is merged **after** caching
```
Toflar: This part, for example, is incorrect. It's not "merged after caching the page". It's rather a placeholder in the original response that is then treated by someone else. In case of ESI it's a reverse proxy. In case of hinclude it's the JS on client-side.

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+    the page, which allows a fragment to have a different cache time than the main page. The `esi` renderer tells
```
Toflar: I think we should just simply say "The `esi` renderer allows fragments to be rendered as Edge Side Includes." and link that to https://en.wikipedia.org/wiki/Edge_Side_Includes. We don't have to explain concepts that are already explained.

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+Before **Contao 4.9**, inline fragments cannot affect the cache time of the page response.
+{{% /notice %}}
+
+By default, fragments like frontend modules and content elements are rendered directly inside the main request.
+This means the content of an inline fragment will be cached for as long as the page is cached, as configured
+in the page setting. If a fragment returns caching information in its response, the page's cache time and the fragment
+will be merged to the lowest common denominator. For example, if the page is cacheable for a day, but the fragment
+only for one hour, the whole page will only be cached for one hour.
+
+A more common use case would be something like a `{{date::Y}}` insert tag, which renders the current year. This
```
Toflar: I don't think it's a good idea to take something as an example which does the exact opposite. `{{date::Y}}` is rendered as ESI at the moment and if I were to read this article I would immediately go and see how this example is technically implemented only to see that it's not the case. That's very confusing.

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+    }
+}
+```
+{{% /expand%}}
+
+
+### Edge Side Includes (ESI)
+
+By setting the fragment renderer to `esi`, the fragment will be cached **separately** from the main content.
+The page can be cached for 24 hours, but the frament can be cached for one week. On every request to the page,
```
Toflar:
```suggestion
The page can be cached for 24 hours, but the fragment can be cached for one week. On every request to the page,
```

Toflar:
There's also quite some spelling issues but I did not comment on them, I think we might fix them later :)

fritzmg:
@aschempp can you update the PR according to @Toflar comments? Everything else can be changed later on.

docs/dev/framework/caching.md
aschempp: Changed in 29eeb9c07ab2b331be81a108d0f16b3a4535af4e

docs/dev/framework/caching.md
aschempp: Obsolete with 29eeb9c07ab2b331be81a108d0f16b3a4535af4e imho. The Wikipedia link is 10x more complicated and technical to understand. We _could_ link to wikipedia, but then again, we can't link to every possible other description on the interwebs?

docs/dev/framework/caching.md
aschempp: I kinda agree, but that would mean we can't provide an example, because the feature is not used by the core yet. I thought about start/stop times of content elements, but that's not implemented either.

docs/dev/framework/caching.md
```diff
@@ -217,24 +217,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+
+
+There are two fundamentally different ways to render fragments:
+ 1. **Inline:** Inside the main page request and merged with the content before caching. In Symfony, this is handled
+    by the `inline` renderer, but Contao provides its own default `forward` renderer, which copies the original
+    request (whereas Symfony does not) to e.g. pass POST data to the subrequest.
+ 2. **Subrequest:** By generating a placeholder for the fragment, the fragment content is rendered **after** caching
+    the page, which allows a fragment to have a different cache time than the main page. In case of `esi` the reverse
+    proxy (cache) will parse these placeholders before delivering to the client.
+    There's also an `hinclude` renderer which allows browser to merge fragments using JavaScript, but the client-side
```
fritzmg:
```suggestion
    There's also an `hinclude` renderer which allows browsers to merge fragments using JavaScript, but the client-side
```

docs/dev/framework/caching.md
Toflar: Then maybe add a hint that this is an example and is still a TODO in Contao core? wdyt @fritzmg

fritzmg: Hmm.. may be create a fictional example with an hypothetical `{{my_date::Y}}` Insert Tag, to not allure to Contao's own `{{date}}` Insert Tag (wich is simply a special case)? I think that would be good, because currently I wouldn't even know how to create an Insert Tag that would be rendered as ESI ;)

Toflar: You can't. It's hardcoded which is why `my_date` wouldn't work :) That's why I'm saying. Andy is taking an example of something we want to do in future but doesn't work at all today.

fritzmg: I see :). Well then may be simply use `MyDateElementController` or something as an example, which simply returns `date('Y')` as the response's content?

docs/dev/framework/caching.md
```diff
@@ -244,24 +244,104 @@ Moreover, you don't have to register to all the different callbacks such as `ons
...
+This means the content of an inline fragment will be cached for as long as the page is cached, as configured
+in the page setting. If a fragment returns caching information in its response, the page's cache time and the fragment
+will be merged to the lowest common denominator. For example, if the page is cacheable for a day, but the fragment
+only for one hour, the whole page will only be cached for one hour.
+
+A more common use case would be something like a `{{date::Y}}` insert tag, which renders the current year. This
+fragment is cacheable until the end of 31st of December. If the page is cached on the 10th of December for one day,
+the page cache time will not be affected. However, if the page is cached at 12pm on the 31st of December, the cache
+time will be lowered to 12 instead of 24 hours. The page cache will expire at the end of the year and regenerated
+on the 1st of January.
```
fritzmg: How about this instead:

```
On the other hand, consider an example of a fragment that renders the current year. This fragment would be cacheable
until the end of 31st of December. If the page is cached on the 10th of December for one day, the page cache time will
not be affected. However, if the page is cached at 12pm on the 31st of December, the cache time will be lowered to 12
instead of 24 hours. The page cache will expire at the end of the year and regenerated on the 1st of January.
```

```php
class CurrentYearController extends AbstractFrontendModuleController
{
    protected function getResponse(Template $template, ModuleModel $model, Request $request): ?Response
    {
        $year = (int) date('Y');
        $template->year = $year;
        $response = $template->getResponse();

        // Cache until the end of the year
        $response->setPublic();
        $response->setMaxAge(strtotime(($year + 1).'-01-01 00:00:00') - time());

        return $response;
    }
}
```

fritzmg: Ping @Toflar @aschempp 

If you agree with this change, I can add it directly to Andreas' PR and then merge this. I'd like to finally get this along, since the PR is almost a year old now 😬

Toflar: Yeah go ahead. We better merge that change. I'll go over the whole article again once it's live and I can see it in full glory ;) just ping me on slack and I'll review again.

fritzmg: Alright :)

fritzmg:
Thank you @aschempp & @Toflar :)

PR #239, opened 2020-01-29
Change second attribute for getAttributesFromDca

fritzmg:
When using `ModulePersonalData` or `ModuleRegistration`, the second parameter can also be of type `\Contao\FrontendModule`.


m-vo (approved):

netzarbeiter (approved):

leofeyer:
The registration module passes `\Contao\Module`, not `\Contao\FrontendModule`. The latter is only used for the service annotation.

fritzmg:
Right, fixed in 5c6f969

PR #241, opened 2020-01-30
Add docs for onrestore_version_callback

dmolineus:
The onrestore_callback is deprecated and the onrestore_version_callback is introduced.


fritzmg:
Can you also adjust https://docs.contao.org/dev/reference/dca/callbacks/ in this PR accordingly?

dmolineus:
> Can you also adjust https://docs.contao.org/dev/reference/dca/callbacks/ in this PR accordingly?

There is already the new callback described. Should the old deprecated one being really described there?

dmolineus:
Oh, it's documented on two places. Didn't saw that. Closed in favour of #243.

PR #243, opened 2020-02-03
Improve callbacks documentation

dmolineus:
This pull request improves the callback documentation

 - Add links to callback page
 - Add deprecated `restore_callback` docs with deprecation note
 - Add missing reference of following callbacks
    - `xlabel`
    - `onundo`

PR #247, opened 2020-02-04
Fix anchor links

dmolineus:
This pull request fixes the broken anchor links introduced by #243. 

Since hugo 0.60 you can configure the markdown parser which also changed the way how headings got converted to ids. Since hugo 0.62.2 there is a new setting `autoHeadingIDType` to preserve the old behaviour.

 - https://gohugo.io/getting-started/configuration-markup/#goldmark
 - https://gohugo.io/news/0.62.2-relnotes/

This PR fixes the anchor ids and sets the new option to preserve the old behaviour.


fritzmg:
Hm, but in your PR you set the old behaviour _and_ you had to change all the anchor links regardless? Ah, or were the previous IDs in your PR already in the new format?

dmolineus:
The IDs in my pr are in the new format (GitHub style which is the default since Hugo 0.60). You're using Hugo 0.58 to build the docs, so my added anchor links are broken at the moment.

I noticed that after the pr got merged and published to docs.contao.org.

I suggest also to update to latest Hugo version when building the docs

fritzmg:
> I suggest also to update to latest Hugo version when building the docs

Yeah, definitely. Will have to test everything though, before updating.

fritzmg (approved):

fritzmg:
Thank you @dmolineus

PR #261, opened 2020-02-14
Documented Contao ME env variables

### LINKED ISSUE #109: Document the environment variables
I just noticed that they're missing and I think we may need to document them in both, the user manual and the dev manual.

As of 4.8, the following variables are relevant:

`APP_ENV`: `prod` (default) enables the production environment and `dev` the development env. It also automatically enables the debug mode and the Symfony "Profiler" (which is about to be renamed apparently :)).

`TRUSTED_PROXIES`: A comma-separated list of trusted proxy IP's.

`TRUSTED_HOSTS`: A comma-separated list of trusted hosts.

`COOKIE_WHITELIST`: A comma-separated list of cookies that should not be stripped in the HTTP cache.

Of course we need a bit more explanation to them all but let's collect them first.

Toflar:
Fixes https://github.com/contao/docs/issues/109

### DIFF
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -3,7 +3,7 @@ title: "Config"
 description: Contao's configuration options.
 ---
 
-### Contao Bundle Configuration
+## Contao Bundle Configuration
 
 Contao's configuration options can be reviewed by running the following command:
 
@@ -144,7 +144,7 @@ contao:
     localconfig:          ~
 ```
 
-### Environment variables for the Contao Managed Edition
+## Environment variables for the Contao Managed Edition
 
 If you use Contao together with the [Contao Managed Edition][Contao_ME], you can use environment variables to influence
 the behaviour of the Managed Edition.
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -3,6 +3,7 @@ title: "Config"
 description: Contao's configuration options.
 ---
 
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

fritzmg:
Thank you @Toflar  :)

fiedsch:
Is setting the environment variable(s) in `.htaccess` with 

```
<IfModule mod_env.c>
  # set env variable(s) for Contao
  setEnv APP_ENV dev
</IfModule>
```
worth mentioning, or is it "obvious"?

Toflar:
That depends on your setup and I think a developer knows how to set it.
In your case, it only works if PHP runs as `mod_php` in Apache. It doesn't work on CLI nor would that work with CGI or FPM :)

fiedsch:
But that kind of proves the point of my question. Does everybody know how to set the variables? 

Maybe links for the different scenarios would be helpful. When reading your text my first reaction was to google "apache set environment variable" which led to my "solution". 

> I think a developer knows how to set it.

I'd make that: a developer should be able to find out (aka. google) how to do it ;-)

Toflar:
I'm not saying everybody knows it. But this is for the developer's manual. We can have a whole section about environment variables and using `.env` as a fallback in the user manual imho :)

fritzmg:
Yeah, this would be more like something for the manual, not the dev docs.

PR #228, opened 2020-01-16
Added docs and examples about fragment controllers

aschempp:
depends on https://github.com/contao/docs/pull/226 for the caching infos

PR #290, opened 2020-02-28
Extend the environment variables with APP_SECRET and DATABASE_URL descriptions

qzminski:
/cc @aschempp

### DIFF
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -172,6 +172,8 @@ contao:
 
 ## Environment variables for the Contao Managed Edition
 
+{{< version "4.9" >}}
+
 If you use Contao together with the [Contao Managed Edition][Contao_ME], you can use environment variables to influence
 the behaviour of the Managed Edition.
 The reason why they are environment variables is because these settings affect the setup before the dependency injection
@@ -225,8 +227,6 @@ container is even built. Settings like trusted proxies or caching are considered
     
 * `COOKIE_WHITELIST`
 
-    {{< version "4.8" >}}
-
     This is a special environment variable related to the default caching proxy which is shipped with the Contao Managed
     Edition by default.
     Contao disables any HTTP caching as soon as there is either a `Cookie` or an `Authorization` header present in the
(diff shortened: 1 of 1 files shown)


fritzmg (approved):
LGTM, though I haven't really tested/used these myself yet (in a Contao Managed Edition - I do use them in regular Symfony applications). I take it they do work in the Contao Managed Edition?

aschempp:
They work in Contao 4.9 and the `parameters.yml` is deprecated in favor of env variables. Both work, but you obviously shouldn't mix them 😆

aschempp (changes requested):
Maybe we should add an info that this only works in 4.9 and one should not define the respective parameters then?

fritzmg:
So just adding `{{< version "4.8" >}}` to `APP_SECRET` and `DATABASE_URL`, is that correct? Plus may be a `{{% notice info %}}` to not mix env and parameters.

aschempp:
Yeah, I guess that's up to you (@fritzmg) 😂 
I didn't use the version tag in my docs, I simply added a note with the same text and some extensive info. Example: https://github.com/contao/docs/pull/228/files#diff-5e7f17e304650b2688078f82ec27a731R49

fritzmg:
Hm, well @Toflar only tagged the `COOKIE_WHITELIST` with `{{< version "4.8" >}}`. Do any environment variables work before 4.8 actually?

aschempp:
Yes, we've added some in 4.6 already: https://github.com/contao/contao/blob/4.6/manager-bundle/src/Resources/skeleton/web/app.php#L34. But before Contao 4.8 you couldn't use the `.env` file (out of the box)

fritzmg:
Ok, well then let's just say that all of these only work in 4.9 :)

aschempp:
I'd be fine with that. And they are the replacement for the parameters.yml in Contao 4.9

fritzmg:
@qzminski can you add these changes?

qzminski:
@fritzmg sure just tell me what and where exactly, as I am not familiar with things like `{{< version "4.8" >}}` 😅

fritzmg:
I have added some changes myself in https://github.com/contao/docs/pull/290/commits/7ae9e76b05131d8afc14a1b68bc7c6602dc43b37 . @aschempp is that about correct?

aschempp (approved):

PR #301, opened 2020-03-10
Add explanation on how to disable indexing for pages conditionally

fritzmg:
This PR adds an explanation on how to dynamically disable search indexing.

@Toflar are there more/better ways? The two ways I described work also in Contao 4.4.

### DIFF
--- a/docs/dev/framework/search-indexing.md
+++ b/docs/dev/framework/search-indexing.md
@@ -186,23 +186,4 @@ Starting with Contao **4.6** you could also fetch the `pageModel` from the reque
 `kernel` event listener.
 
 
-## Ignoring GET Parameters
-
-The search index creates entries for each unique URL, including its GET parameters, since those parameters typically
-change the content of the page. However, you might want to ignore URLs with certain GET parameters to be added to the
-search index. Contao provides a legacy way of doing this via the `$GLOBALS['TL_NOINDEX_KEYS']` global array. It already
-contains the following default GET parameters:
-
-```php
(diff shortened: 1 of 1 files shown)


docs/dev/framework/search-indexing.md
```diff
@@ -148,3 +150,59 @@ $jsonLdScriptsData =  $document->extractJsonLdScripts('https://schema.org', 'Pro
...
+        }
+    }
+}
+```
+
+Starting with Contao **4.6** you could also fetch the `pageModel` from the request attributes (if present) in your own 
+`kernel` event listener.
+
+
+## Ignoring GET Parameters
```
Toflar: I wouldn't document that because it's actually just plain wrong that this even works.
It will get indexed which is correct. If you don't want this to happen, just set the page to `noSearch` or `noindex` depending on the GET parameters set.

fritzmg: Well, if you have a page `https://example.org/foo.html` and a bot calls `https://example.org/foo.html?foobar=foo`, then you will get two search results. Not sure what the best way to deal with that is. I was merely documenting what already works.

Toflar: Yeah, and what I'm saying is that just because it works, doesn't mean it should.
We cannot remove that because of BC but you should certainly not use it.

You will always get multiple cache entries. It's easy to flood caches by just adding nonsense query parameters to URLs but that's just how caches work.

If you want to send a 404 specifically when `foobar=foo` does not exist in your module, you can.

fritzmg: It's now removed :)

PR #303, opened 2020-03-18
added file path for Contao <4.8

Teetrinker:
`app/Resources/contao/dca/`
vs.
`contao/dca/`

### DIFF
--- a/docs/dev/getting-started/dca.md
+++ b/docs/dev/getting-started/dca.md
@@ -10,7 +10,11 @@ be adjusted to quickly add new fields to content elements, front end modules & n
 entries etc., change some of their properties for your specific needs and to set up
 new database tables for your own custom records.
 
-Any DCA adjustments go into the `contao/dca/` folder. There, you have to create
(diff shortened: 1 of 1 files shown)


fritzmg:
That's not necessary. The difference between Contao 4.4 and 4.9 is explained here: https://docs.contao.org/dev/getting-started/starting-development/#structure

Teetrinker:
I didn't find that information when reading the doc. While reading about the dca, I had no clue paths could differ.

I don't know the best way to do this, but this little notice about differing paths should be more accessible for people diving straight in.

fritzmg:
Well, people diving straight in should may be also use the most recent Contao version ;). The `contao/` path is referenced in many, many places obviously, more than you have covered in your PR. Always referencing

```
app/Resources/contao/
```

and

```
contao/
```

seems superfluous to me.

We could of course _only_ reference `app/Resources/contao/`, since that will aways work - but we want people to use the best practice approach, not the legacy approach.

_But_, may be a short notice in these specific places are warranted.

fritzmg:
I have now added more references to the folder structures in various places in 5a44c26

Issue #300, opened 2020-03-09
Update Symfony Application configuration for Contao 4.9


fritzmg:
@sheeep could you please provide PR to update https://docs.contao.org/dev/getting-started/starting-development/ to include the necessary changes for Contao 4.9? :)

May be split it between Contao 4.4 or 4.9? Both should stay imho.

bytehead:
You mean over here: https://docs.contao.org/dev/getting-started/initial-setup/symfony-application/?

fritzmg:
Right sorry, copied the wrong link.

bytehead:
Can you assign the issue to me then?

fritzmg:
Done.

bytehead:
@fritzmg this issue can be closed.

PR #307, opened 2020-03-22
Split Contao 4.4.* / 4.9* in a Symfony application docs into separate pages

bytehead:
See #300.

Todo:

- [x] Contao 4.4 LTS
- [x] Contao 4.9 LTS

### DIFF
--- a/docs/dev/getting-started/initial-setup/symfony-application/_index.md
+++ b/docs/dev/getting-started/initial-setup/symfony-application/_index.md
@@ -1,0 +1,11 @@
+---
+title: "Use in Symfony Application"
+description: "How to integrate Contao into a Symfony application."
+aliases:
+  - /getting-started/initial-setup/symfony-application/
+---
+
+
+Choose your Contao version:
+
+{{% children description="true" %}}
--- a/docs/dev/getting-started/initial-setup/symfony-application/contao-4.4.md
+++ b/docs/dev/getting-started/initial-setup/symfony-application/contao-4.4.md
@@ -64,7 +64,7 @@ and `doctrine/migrations` need to be installed in another version than the defau
 one.
 
 As long as the Symfony flex plugin is installed you will be asked to execute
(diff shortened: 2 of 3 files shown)


fritzmg (changes requested):
`contao-4.9.md` still mentions the following:

> And change the default and fallback language to `de`, so the install tool comes
up translated. In order to do so, change `en` to `de` in `config/translation.yaml`.

But as far as I understood it, in your tutorial, the Install Tool isn't used anymore (nor is it needed, right?). So may be this paragraph needs to be rewritten slightly.

fritzmg:
Btw. there is a PR for tabs for the Learn Theme: https://github.com/matcornic/hugo-theme-learn/pull/386

So may be in the future we can consolidate it into one article again.

bytehead:
> But as far as I understood it, in your tutorial, the Install Tool isn't used anymore (nor is it needed, right?). So may be this paragraph needs to be rewritten slightly.

It's still with the use of the install tool as you cannot create an admin user on the CLI (see https://github.com/contao/contao/pull/1463). I'll update the docs accordingly, after the command is available.

See https://github.com/contao/docs/pull/307/files#diff-987fb580f1f4826e15ddf477dedff4eaR256

bytehead:
Yeah, let me know when the tab feature is available! 🎉

fritzmg:
> It's still with the use of the install tool as you cannot create an admin user on the CLI (see [contao/contao#1463](https://github.com/contao/contao/pull/1463)). I'll update the docs accordingly, after the command is available.

Ah yes, but in the 4.9 version, you don't add the Install Tool's routes to the `routes.yaml`. This has still to be done, or not?

bytehead:
Yes, good catch. I've updated it.

fritzmg (approved):

fritzmg:
Thank you @bytehead :)

Issue #310, opened 2020-03-23
Document ContaoModuleBundle


fritzmg:
If you install packages of type `contao-module` in your Symfony application, you will need to use the `ContaoModuleBundle` class in order to load these extensions in the application.

@bytehead can you add an article about that to the Symfony Application setup documentation? :)

bytehead:
👍

fritzmg:
Btw. @ausi suggested [here](https://github.com/contao/contao/issues/216#issuecomment-447897985) that one could may be reuse `ContaoKernel::registerBundles` to automatically load any `contao-module` packages?

bytehead:
Should it also be possible to define the environments the modules gets loaded, like e.g.:
```php
return [
    'easy_themes' => ['all' => true],
];
```

or should the system always load `contao-module` packages?

fritzmg:
I have no idea 😁

bytehead:
Loading them in every environment will be automatically possible without manually adding the module somewhere after a `composer require`...

bytehead:
This can be closed.

PR #313, opened 2020-03-24
Load Contao modules automatically, if available

bytehead:
See #310.

@fritzmg WDYT?


fritzmg:
lgtm :). May be add a note about what that new code actually does? Doesn't need to go into detail, just that it would automatically load any Contao 3 extensions, if installed.

bytehead:
> May be add a note about what that new code actually does? Doesn't need to go into detail, just that it would automatically load any Contao 3 extensions, if installed.

Sure, it was more like a proposal 🙃

bytehead:
Updated in 2c44afc.

fritzmg (approved):

fritzmg:
Thank you @bytehead

PR #322, opened 2020-03-29
Document Contao's events

### LINKED ISSUE #18: Reference: Events

fritzmg:
Closes #18

@Toflar could you have a look at the description of `contao.robots_txt`? Looking at the code I wasn't actually sure how one could add custom records to the `robots.txt`.

### DIFF
--- a/docs/dev/guides/back-end-routes.md
+++ b/docs/dev/guides/back-end-routes.md
@@ -3,6 +3,7 @@ title: "Back End Routes"
 description: "Custom routes within the Contao back end."
 aliases:
   - /guides/backend-routes
+  - /guides/back-end-routes
 ---
 
 
@@ -145,7 +146,7 @@ class BackendMenuListener
         $factory = $event->getFactory();
         $tree = $event->getTree();
 
-        if ('mainMenu' !== $tree->getName() ) {
+        if ('mainMenu' !== $tree->getName()) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


Toflar:
That looks correct to me. Well you may link to webignition because that‘s the library used. We use it in the core to add the sitemap URL if you need an example :)

fritzmg:
@Toflar Hm, to be honest, may be I don't get how `webignition/robots-txt-file` is supposed to work. As mentioned in https://github.com/contao/contao/pull/1590, I always end up with a `robots.txt` that looks like this:
```
user-agent:*
disallow:/foo/
disallow:/contao/

user-agent:*
disallow:/contao/
```

instead of this:

```
user-agent:*
disallow:/foo/
disallow:/contao/
```

when my own event listener is executed before Contao's `RobotsTxtListener` (which happens by default without priorities). May be there is nothing wrong with the implementation in Contao after all, but if so, can you give me an example, that will lead to

``` 
user-agent:*
disallow:/foo/
disallow:/contao/
```

?

fritzmg:
nvm. got it now, see https://github.com/contao/contao/pull/1590#issuecomment-605708138 :)

I have updated this PR with a an updated example as well.

fritzmg:
I'll merge this for now. I hope other people can provide examples for other events at some point ;).

fiedsch (commented):
https://github.com/contao/docs/pull/322/files#diff-2927ffa9bc82d51b4a0e587372efd363R188

This sentences misses something like „is created“

fritzmg:
Fixed in 93e5f30f61f2ee07dcea9ab99461e0400d66182d

PR #325, opened 2020-04-01
Update request-tokens.md

Mynyx:
See https://github.com/contao/contao/blob/bf87609767e5623d9da8b98920683bef2d651997/core-bundle/src/EventListener/RequestTokenListener.php#L71-L75

### DIFF
--- a/docs/dev/framework/request-tokens.md
+++ b/docs/dev/framework/request-tokens.md
@@ -20,8 +20,9 @@ browser. Only if you are authenticated in such a way that the browser will autom
 along without any user interaction (logged in via `Cookie` or basic authentication  via `Authorization` headers) or the
 request has any other cookies, CSRF protection is required. So don't get fooled by the cookies not being present all the time, Contao is actually very smart about them to improve HTTP cache hits.
 
-By default, Contao protects all `POST` requests except Ajax requests coming from Contao routes. That means, routes that either do have
-the Route attribute `_scope` set to `frontend` or `backend`.
+By default, Contao protects all `POST` requests except Ajax requests coming from Contao routes. That means, routes that
+either do have the Route attribute `_scope` set to `frontend` or `backend`. (Ajax requests cannot be used for CSRF attacks,
+because they are secured against cross domain requests by the browser.)
 
 If you explicitly want to disable the CSRF protection on your own route, you can set the route attribute `_token_check`
 to `false`.
--- a/docs/dev/framework/request-tokens.md
+++ b/docs/dev/framework/request-tokens.md
@@ -21,9 +21,8 @@ along without any user interaction (e.g. any cookies or basic authentication via
 protection is required. So don't get fooled by the cookies not being present all the time, Contao is actually very smart
 about them to improve HTTP cache hits.
 
-By default, Contao protects all `POST` requests except Ajax requests coming from Contao routes. That means, routes that
-either do have the Route attribute `_scope` set to `frontend` or `backend`. (Ajax requests cannot be used for CSRF attacks,
-because they are secured against cross domain requests by the browser.)
+By default, Contao protects all `POST` requests coming from Contao routes, except Ajax requests. That means, routes that
+either do have the Route attribute `_scope` set to `frontend` or `backend`.
 
 If you explicitly want to disable the CSRF protection on your own route, you can set the route attribute `_token_check`
 to `false`.
--- a/docs/dev/framework/request-tokens.md
+++ b/docs/dev/framework/request-tokens.md
@@ -21,8 +21,8 @@ along without any user interaction (e.g. any cookies or basic authentication via
 protection is required. So don't get fooled by the cookies not being present all the time, Contao is actually very smart
 about them to improve HTTP cache hits.
 
-By default, Contao protects all `POST` requests coming from Contao routes, except Ajax requests. That means, routes that
-either do have the Route attribute `_scope` set to `frontend` or `backend`.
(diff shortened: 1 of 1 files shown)


fritzmg:
@ausi hm, why does the request token need to be checked if there is _any_ cookie in the request?

Toflar:
Because any cookie might be used to personalize content. But I don't understand this PR. It already says that if cookies are sent?

`any user interaction (logged in via `Cookie` or basic authentication  via `Authorization` headers)`

That is the definition of `any cookie` to me.

fritzmg:
I'd write the following then:

Before:

```diff
-Only if you are authenticated in such a way that the browser will automatically send authentication information along without any user interaction (logged in via `Cookie` or basic authentication  via `Authorization` headers)
+Only if you are authenticated in such a way that the browser will automatically send authentication information along without any user interaction (e.g. any cookies or basic authentication  via `Authorization` headers)
```

The 

> logged in via `Cookie`

makes it sound like it is indeed only necessary with an authentication cookie.

Toflar:
Agreed. It's any cookie that might identify you personally.
`$response->headers->setCookie(Cookie::create('username', 'foobar'));` for example.

fritzmg:
Well, you mean any cookie _could_ potentially personally identify know, but there is know way of knowing which cookie does - so the request token has to be checked on every request, even though none of the cookies might contain personal information, correct?

Toflar:
That's correct. But cookies are explicitly meant to identify someone personally, right? They allow you to track someone across requests which is the epitome of "identifying someone" :)
Maybe you track the last page the user visited or the last button they clicked. It's all considered personal information.
If you use cookies to track information that is probably not related to a user at all and would not require a CSRF protection (like e.g. the current time or whatever) then you probably just don't need cookies anyway :)

fritzmg:
> But cookies are explicitly meant to identify someone personally, right?

Meant to, but not always used to ;).

Toflar:
Yeah. Then things just don't work out as expected and developers have to finally read and understand the HTTP specs. We're doing something good to the world here :)

docs/dev/framework/request-tokens.md
```diff
@@ -17,11 +17,10 @@ The way Contao mitigates these attacks is by using the [Double Submit Cookie tec
 Providing CSRF protection for users that are not authenticated against the app does not make any sense. So if you
 visit a regular Contao page with a form placed on it, you will not necessarily see any cookies being set in your
 browser. Only if you are authenticated in such a way that the browser will automatically send authentication information
-along without any user interaction (logged in via `Cookie` or basic authentication  via `Authorization` headers), CSRF
-protection is required. So don't get fooled by the cookies not being present all the time, Contao is actually very smart
-about them to improve HTTP cache hits.
+along without any user interaction (logged in via `Cookie` or basic authentication  via `Authorization` headers) or the
+request has any other cookies, CSRF protection is required. So don't get fooled by the cookies not being present all the time, Contao is actually very smart about them to improve HTTP cache hits.
```
fritzmg:
```suggestion
along without any user interaction (e.g. any cookies or basic authentication via `Authorization` headers), CSRF
protection is required. So don't get fooled by the cookies not being present all the time, Contao is actually very smart
about them to improve HTTP cache hits.
```

fritzmg (changes requested):

docs/dev/framework/request-tokens.md
```diff
@@ -17,11 +17,10 @@ The way Contao mitigates these attacks is by using the [Double Submit Cookie tec
...
 
-By default, Contao protects all `POST` requests coming from Contao routes. That means, routes that either do have
+By default, Contao protects all `POST` requests except Ajax requests coming from Contao routes. That means, routes that either do have
```
fritzmg: Hm, this I find confusing too. Why is _any_ AJAX request not checked? Shouldn't any `POST` request being checked, regardless of whether it is identified as an AJAX request or not?

ausi: Ajax requests cannot be used for CSRF attacks, because they are secured against cross domain requests by the browser.

ausi: To be more precise: We check for AJAX requests by validating that the HTTP header `X-Requested-With` was set to `XMLHttpRequest`, see https://github.com/symfony/symfony/blob/e1a522b47077b7673090912e33469ac9f0b72a67/src/Symfony/Component/HttpFoundation/Request.php#L1712

Setting custom HTTP headers is only allowed for AJAX requests on the same origin and doesn’t work cross-domain, except if it was explicitly allowed by a CORS preflight request. For details see <https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#use-of-custom-request-headers>

Toflar: I think in this PR we should just omit any explanation and just write we don't check request tokens for ajax requests. We can include the explanation for it in an info notice. I'll do a PR if @Mynyx doesn't have the energy anymore :)

Mynyx: @Toflar A PR would be welcome. :)

Toflar: Sure, I'll do that once this PR is ready. Please remove any explanation and incorporate other reviews so we can finish this :)

fritzmg:
@Mynyx can you update this?

fritzmg:
I have merged this with some additional changes, thank you @Mynyx :).

@Toflar feel free to create a follow up PR for the Ajax request clarification.

PR #327, opened 2020-04-02
Update config.md

Mynyx:
Add ``trusted_device`` cookie to ``COOKIE_WHITELIST``.


fritzmg:
Based on the discussion in https://github.com/contao/contao/issues/1618

Mynyx:
Currently, the trusted devices function wouldn't work if the cookie is not included in the whitelist.

**Edit:** Of course, this only applies if the whitelist is used at all.

fritzmg:
> Currently, the trusted devices function wouldn't work if the cookie is not included in the whitelist.

This is a comment you should add in https://github.com/contao/contao/issues/1618, not here.

Mynyx:
@Toflar I think this change would correctly describe the current situation?

Toflar (approved):
That's correct.

fritzmg (approved):

fritzmg:
Thank you @Mynyx :)

Issue #31, opened 2019-05-22
Documentation: asset helper


fritzmg:
https://community.contao.org/de/showthread.php?74997-Webpack-Encore&p=504390&viewfull=1#post504390

fkaminski:
Is there still a need? Because you already did it:
https://docs.contao.org/dev/framework/asset-management/

fritzmg:
Yeah, no, forgot to close this issue ;)

PR #332, opened 2020-04-04
Asset Management

fritzmg:
Implements #31

### DIFF
--- a/docs/dev/framework/asset-management.md
+++ b/docs/dev/framework/asset-management.md
@@ -50,6 +50,156 @@ $GLOBALS['TL_CSS'][] = 'bundles/myextension/frontend.css';
 Contao allows a few options to be set for each file. These options are appended
 to the file path and separated by a pipe `|` character.
 
+By appending `|static` you can define the stylesheet as being "static", meaning 
+that it can be combined with other static stylesheets into one file (if enabled
+in Contao's page layout). Without the `static` option, the stylesheet is always 
+included separately. A stylesheet should be defined as `static`, if it will occur 
+on every or most pages of a website, as it is more advantageous then for it to be 
+combined with other static stylesheets.
+
+```php
+$GLOBALS['TL_CSS'][] = 'bundles/myextension/frontend.css|static';
+```
+
+In case of stylesheets, you can also define its _media_ property the same ways as
+defining it as "static". For instance, if a stylesheet is only to be included for
+the `print` media, append `|print` to the stylesheet path.
+
+```php
+$GLOBALS['TL_CSS'][] = 'files/theme/css/print.css|print';
+```
+
+This results in the following HTML code in the front end:
+
+```html
+<link rel="stylesheet" href="files/theme/css/print.css?v=da0d4373" media="print">
+```
+
+Both options can be combined, in no particular order:
+
+```php
+$GLOBALS['TL_CSS'][] = 'files/theme/css/print.css|print|static';
+```
+
+
+### Adding JavaScript
+
+Adding JavaScript to the `<head>` of a document works the same ways as [adding CSS](#adding-css).
+This time, the paths to the script files are added to the `$GLOBALS['TL_JAVASCRIPT']`
+array.
+
+```php
+$GLOBALS['TL_JAVASCRIPT'][] = 'bundles/myextension/scripts.js';
+```
+
+As with CSS, JavaScript assets in the `<head>` of a document can be combined with
+others (if enabled in the page layout) if they are defined as "static":
+
+```php
+$GLOBALS['TL_JAVASCRIPT'][] = 'bundles/myextension/scripts.js|static';
+```
+
+There are no further options to be defined for JavaScript assets.
+
+
+### Generating Style and Script Tags
+
+As mentioned in the introduction, it is also possible to add custom HTML to either
+the `<head>` or the end of the `<body>` element. Often this is used to add assets
+manually, by adding the apropriate `<link>`, `<style>` or `<script>` HTML tags.
+
+Contao offers a few static utility functions for this purpose via the `\Contao\Template`
+class.
+
+* __`\Contao\Template::generateStyleTag($href, $media, $mtime)`__ 
+  
+  This returns a `<link rel="stylesheet" …>` tag and takes three arguments: the
+  path to the stylesheet (absolute or relative to the _base_), an optional _media_
+  attribute and an optional modification time, which Contao will use to append a
+  query parameter to the file for cache busting. The latter can also be set to
+  `null` in order to automatically use the file's modification time for cache busting
+  (if it is a relative file path).
+
+  ```php
+  $GLOBALS['TL_HEAD'][] = \Contao\Template::generateStyleTag('bundles/myextension/print.css', 'print', null);
+  ```
+* __`\Contao\Template::generateInlineStyle($script)`__
+
+  This wraps the given CSS with `<style>…</style>`.
+
+  ```php
+  $GLOBALS['TL_HEAD'][] = \Contao\Template::generateInlineStyle($this->generateCss());
+  ```
+* __`\Contao\Template::generateScriptTag($href, $async, $mtime, $hash, $crossorigin, $referrerpolicy)`__ 
+  
+  This returns a `<script src="…" …>` tag and takes six arguments: 
+
+  * `$href`: the path to the stylesheet (absolute or relative to the _base_)
+  * `$async`: whether the `async` attribute should be added to the tag (default `false`)
+  * `$mtime`: an optional modification time, which Contao will use to append a query 
+    parameter to the file for cache busting. This can also be set to `null` in order 
+    to automatically use the file's modification time for cache busting.
+  * `$hash`: optional hash for an `integrity` attribute.
+  * `$crossorigin`: optional `crossorigin` attribute.
+  * `$referrerpolicy`: optional `referrerpolicy` attribute.
+
+  {{% notice note %}}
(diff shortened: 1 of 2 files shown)


docs/dev/framework/asset-management.md
```diff
@@ -0,0 +1,205 @@
...
+
+Contao utilises global arrays to store references for later inclusion into the page
+layout.
+
+
+| Array | Description |
+| --- | --- |
+| `$GLOBALS['TL_BODY']` | Contains HTML code to be included before `</body>`.<sup>1</sup> |
+| `$GLOBALS['TL_CSS']` | Contains relative paths to CSS assets to be included in the `<head>` of the document. |
+| `$GLOBALS['TL_FRAMEWORK_CSS']` | Contains relative paths to CSS assets to be included in the `<head>` of the document. |
```
aschempp: not sure if we should document that as it shouldn't be used by devs?

docs/dev/framework/asset-management.md
```diff
@@ -0,0 +1,205 @@
...
+| `$GLOBALS['TL_HEAD']` | Contains HTML code to be included in the `<head>` of the document. |
+| `$GLOBALS['TL_JAVASCRIPT']` | Contains relative paths to JavaScripts assets to be included in the `<head>` of the document. |
+| `$GLOBALS['TL_JQUERY']` | Contains HTML code to be included before `</body>`.<sup>1</sup> |
+| `$GLOBALS['TL_MOOTOOLS']` | Contains HTML code to be included before `</body>`.<sup>1</sup> |
+| `$GLOBALS['TL_USER_CSS']` | Contains relative paths to CSS assets to be included in the `<head>` of the document. |
```
aschempp: path can be absolute as well

docs/dev/framework/asset-management.md
```diff
@@ -0,0 +1,205 @@
...
+
+
+| Array | Description |
+| --- | --- |
+| `$GLOBALS['TL_BODY']` | Contains HTML code to be included before `</body>`.<sup>1</sup> |
+| `$GLOBALS['TL_CSS']` | Contains relative paths to CSS assets to be included in the `<head>` of the document. |
+| `$GLOBALS['TL_FRAMEWORK_CSS']` | Contains relative paths to CSS assets to be included in the `<head>` of the document. |
+| `$GLOBALS['TL_HEAD']` | Contains HTML code to be included in the `<head>` of the document. |
+| `$GLOBALS['TL_JAVASCRIPT']` | Contains relative paths to JavaScripts assets to be included in the `<head>` of the document. |
+| `$GLOBALS['TL_JQUERY']` | Contains HTML code to be included before `</body>`.<sup>1</sup> |
```
aschempp: jquery and mootools are only included if the respective checkbox is enabled in the layout, as far as I remember?

docs/dev/framework/asset-management.md
```diff
@@ -0,0 +1,205 @@
...
+```
+
+As with CSS, JavaScript assets in the `<head>` of a document can be combined with
+others (if enabled in the page layout) if they are defined as "static":
+
+```php
+$GLOBALS['TL_JAVASCRIPT'][] = 'bundles/myextension/scripts.js|static';
+```
+
+There are no further options to be defined for JavaScript assets.
```
aschempp: isn't there `async` as well?

docs/dev/framework/asset-management.md
```diff
@@ -0,0 +1,205 @@
...
+
+For example, in order to integrate the tablesorter component (which Contao installs
+as a dependency through `contao-components/tablesort`) within a template, the following
+can be used:
+
+```php
+<script src="<?= $this->asset('js/tablesort.min.js', 'contao-components/tablesort') ?>"></script>
+```
+
+The same can be used for assets of a bundle. Suppose you have a package called `example/contao-foo-extension`
```
aschempp: I'm not sure this is correct. If I remember correctly, bundles use the bundle name, not the package name, as each package can contain multiple bundles. We only use the package name for `contao-component`s. Let me know if you need help digging that in the Contao code 🙈

docs/dev/framework/asset-management.md
fritzmg: Yeah, may be not.

docs/dev/framework/asset-management.md
fritzmg: No, that's always processed. Nothing will be _in_ those global arrays, if you haven't activated any templates for it - but if you add something to those arrays yourself, it will be shown in the front end.

docs/dev/framework/asset-management.md
fritzmg: Oh, yeah :)

docs/dev/framework/asset-management.md
fritzmg: To be honest, I did not test this my self 😁 . But the parameter is called `$packageName`, not `$bundleName`, and for the Contao Components, you also use the package name. But looking at `AddAssetsPackagesPass::addBundles`, it seems to be indeed the bundle name.

docs/dev/framework/asset-management.md
fritzmg: I have removed `TL_JQUERY` nonetheless.

ausi (commented):
It is also possible to add a pipe followed by a numeric string to set the `mtime` option, e.g. `|10184084`. See https://github.com/contao/contao/blob/2560349f1985eae6eb69da4f2eeb9a3ad6e4f37a/core-bundle/src/Resources/contao/library/Contao/StringUtil.php#L858

`mtime` doesn’t have to be a modification time AFAIK, it can be used for a numeric version too.

docs/dev/framework/asset-management.md
```diff
@@ -0,0 +1,211 @@
+---
+title: "Asset Management"
+description: Managing assets (JavaSCript, CSS) in the front end.
```
Toflar:
```suggestion
description: Managing assets (JavaScript, CSS) in the front end.
```

Toflar (commented):
I think the `{{asset::*}}` insert tag which works with the `manifest.json` might good to mention too. /cc @aschempp

fritzmg:
@ausi True. Originally I deemed it not worth mentioning, since I thought Contao would automatically add the version parameter for any such assets. But that's not actually the case.

fritzmg:
@Toflar good point. As far as I can see the `{{asset::*}}` Insert Tag works the same way as `Template::asset()`. However, I have no clue and experience about its role with the `manifest.json`.

Toflar:
That's where @aschempp might provide some insights :)

aschempp:
> As far as I can see the `{{asset::*}}` Insert Tag works the same way as `Template::asset()`. 

correct

> However, I have no clue and experience about its role with the `manifest.json`.

The `manifest.json` works for all assets, not just the insert tag. 

---

For the basics: We're using the [The Symfony Asset component](https://symfony.com/doc/current/components/asset.html) here. From the docs:

 > The Asset component manages URL generation and versioning of web assets such as CSS stylesheets, JavaScript files and image files.

Assets are grouped by "package", each package can have a different version strategy. The template method and insert tag are just a shortcut to the framework asset component, so Contao supports whatever Symfony Assets supports.

There are several use cases in Contao, some because Contao automatically registers asset packages:

 1. Packages of type `contao-component` are registered by their package name. Therefore, one can include the (imaginary) file in `/assets/jquery/jquery.js` through `{{asset::jquery.js::contao-component/jquery}}`. It uses the package version to generate a strategy that adds unique file names, so the resulting path for `contao-component/jquery: 1.1.0` will be `/assets/jquery/jquery.js?v=1.1.0`.

 2. If a bundle has a `/public` folder (or it's deprecated sibling `src/Resources/public`), that folder is registered as asset through the bundle short name. A (imaginary) file in `/vendor/contao/news-bundle/Resources/public/foobar.js` is symlinked by Symfony to `/web/bundles/contao_news/foobar.js` (regardless of the asset component).

    - By default, a bundle does not have a version, so no version suffix is added. The asset component resoves `{{asset::foobar.js::contao_news}}` to `/web/bundles/contao_news/foobar.js`.
    - If the `public` folder has a `manifest.json` in it's root, that file is used to generate a *manifest.json version strategy*. For `{{asset::foobar.js::contao_news}}` the asset component will look up `foobar.json` in the `manifest.json` and output that path. This is useful when using e.g. Webpack Encore to generate the assets, as hashed file names are resolved to `/web/bundles/contao_news/foobar-1ussdg71.js`.

fritzmg:
I have now restructured the whole article and overhauled the _Accessing Assets in Templates_ section.

docs/dev/framework/asset-management.md
```diff
@@ -29,30 +29,57 @@ the end of the `<body>`.
...
-that it can be combined with other static stylesheets into one file (if enabled
-in Contao's page layout). Without the `static` option, the stylesheet is always 
-included separately. A stylesheet should be defined as `static`, if it will occur 
-on every or most pages of a website, as it is more advantageous then for it to be 
-combined with other static stylesheets.
+| Option | Example | Description |
+| --- | --- | --- |
+| Static | `|static` | Defines the asset as "static". |
+| Media | `|print` | Defines the `media` attribute of the `<link>` tag (CSS only). |
+| Media | `|asnyc` | Defines the `async` attribute of the `<script>` tag (JavaScript only). |
```
xchs, line 55:
```suggestion
| Media | `|async` | Defines the `async` attribute of the `<script>` tag (JavaScript only). |
```

ausi (approved):

PR #367, opened 2020-05-22
Documented the contao:crawl command

### LINKED ISSUE #296: Reference - Commands: contao:crawl command
plus all available options (e.g. custom Headers)

Toflar:
Closes https://github.com/contao/docs/issues/296


docs/dev/reference/commands.md
```diff
@@ -70,6 +70,47 @@ ask for it. The following tasks are available.
...
+* `broken-link-checker` - checks all the pages for broken links
+
+Any extension might provide additional subscribers so this list is not necessarily conclusive.
+Use it as follows:
+
+```sh
+$> php bin/console contao:crawl [options] [<job>]
+```
+
+There's only one argument to this command which is `job`. It is optional and represents a job ID.
```
fritzmg:
```suggestion
There is only one argument to this command which is `job`. It is optional and represents a job ID.
```

docs/dev/reference/commands.md
```diff
@@ -70,6 +70,47 @@ ask for it. The following tasks are available.
...
+Crawling can be a very long-running task so in case you want to stop and pick up where you left later,
+you need to remember the job ID you were assigned when you first ran the command and you can then
+resume with that job ID later on. However, as there's usually no such thing as memory or runtime limits
```
fritzmg:
```suggestion
resume with that job ID later on. However, as there is usually no such thing as memory or runtime limits
```

docs/dev/reference/commands.md
```diff
@@ -70,6 +70,47 @@ ask for it. The following tasks are available.
...
+on CLI, this will likely not be used very often.
+
+The options are far more important so let's get to them right away:
+
+
+| Option               | Description |
+|----------------------|-------------|
+| `--subscribers (-s)` | By default, all subscribers are enabled but you can pass a comma-separated list of subscribers you want to enable. E.g. if you only want to check for broken links, only pass `broken-link-checker` here. |
+| `--concurrency (-c)` | This option allows to configure the number of concurrent requests. The higher, the faster the process will complete but webservers only handle a certain amount of concurrent requests so choose the value wisely. |
+| `--delay`            | To make sure you're not hammering a webserver, you can also configure a delay. This is in microseconds and will just cause the crawler to wait `n` microseconds between requests. |
```
fritzmg:
```suggestion
| `--delay`            | To make sure you are not hammering a webserver, you can also configure a delay. This is in microseconds and will just cause the crawler to wait `n` microseconds between requests. |
```

docs/dev/reference/commands.md
```diff
@@ -70,6 +70,47 @@ ask for it. The following tasks are available.
...
+| `--max-requests`     | By default, there's no limit configured but if you want to only execute a certain number of requests in total, you can do that with this option. If you want to pick up the job later, see the `job` argument. |
```
fritzmg:
```suggestion
| `--max-requests`     | By default, there is no limit configured but if you want to only execute a certain number of requests in total, you can do that with this option. If you want to pick up the job later, see the `job` argument. |
```

docs/dev/reference/commands.md
```diff
@@ -70,6 +70,47 @@ ask for it. The following tasks are available.
...
+| `--max-depth`        | This is the tree depth, the crawler is going to search. The root page is basically level 1 and all the links found there will be level 2. All the links found on level 2 will be level 3 and so on. By default, there's no max-depth configured. The higher the number, the deeper the crawler will search but it will also take longer. |
```
fritzmg:
```suggestion
| `--max-depth`        | This is the tree depth, the crawler is going to search. The root page is basically level 1 and all the links found there will be level 2. All the links found on level 2 will be level 3 and so on. By default, there is no max-depth configured. The higher the number, the deeper the crawler will search but it will also take longer. |
```

fritzmg (changes requested):
Looks great :). Only nitpick is the usage of contractions which I think we should refrain from using in the documentation.

fritzmg (approved):

fritzmg:
Thank you @Toflar

PR #376, opened 2020-05-25
Make all hook params consistent

Tastaturberuf:


fritzmg:
I would like to stay consistent. In one hook we now have `ModuleModel $model` and `Module $module`.

So here it should either be `PageModel $model` and `PageRegular $page` or `PageModel $pageModel` and `PageRegular $pageRegular`.

Tastaturberuf:
Ah i know what you mean. i will check all hooks and push it in this PR. My goal is to be consistent too.

docs/dev/reference/hooks/modifyFrontendPage.md
```diff
@@ -25,7 +25,7 @@ insert tags, use the [`outputFrontendTemplate`](../outputFrontendTemplate) hook
 
     Content of the rendered front end template.
 
-2. *string* `$templateName`
+2. *string* `$template`
```
fritzmg, line 28: Not sure about this. In the `parseTemplate` hook for example, there is a *Template* `$template`. I think that's why I called it `$templateName` here, since it is just the template name and not the `Template` instance.

Tastaturberuf: but two things:

1. everyone see it is string
2. https://docs.contao.org/dev/reference/hooks/outputFrontendTemplate/
https://docs.contao.org/dev/reference/hooks/outputBackendTemplate/
https://docs.contao.org/dev/reference/hooks/parseFrontendTemplate/

Tastaturberuf: i think when you can see the type, it is clear that is not an object.

fritzmg: I still think the naming is wrong, since it's not the template itself (neither the `Template` instance, nor the template _content_) but only the name of the template.

Tastaturberuf: otherwise we must name all string templates `$templateName`

fritzmg: Yes, that's what should be done :).

Tastaturberuf: and here https://docs.contao.org/dev/reference/hooks/postDownload/ `$fileName`?

fritzmg: Hm. I guess so. I mean, I think it is more important for `$template` vs. `$templateName`, because `string $template` could be misconstrued as the _contents_ of a template within the context of the previously mentioned hooks. At least I wouldn't call it `string $template` in my own code 😁

Tastaturberuf:
Here is an overview with all Contao hooks: https://gist.github.com/Tastaturberuf/8851fc3348dcc229f4a49d81bc9220d0

On first sign i think we should suffix all models with `Model` to make clear what it is.

e.g.
https://docs.contao.org/dev/reference/hooks/getPageLayout/
https://docs.contao.org/dev/reference/hooks/getFrontendModule/

fritzmg:
@Tastaturberuf not sure you'd still be working on this - but: in Contao itself we usually use just `$page` for a `PageModel` for example. So I think we should do the same in the docs (I have adjusted the `getPageLayout` hook accordingly).

Regarding the `getFrontendModule` case: there I think it makes sense to just use `$model` and `$module` as this hook specifically handles modules and thus you only have to distinguish this way.

Tastaturberuf:
I'am not working active on this issue. It was a time i have introduces [Contao interfaces](https://github.com/Tastaturberuf/contao-interfaces/tree/master/Hooks) to have a better experience with hooks and not to open the documentation every time. Maybe i'm merge this library with https://github.com/Tastaturberuf/contao-datacontainer-bundle and give it a new shot.

fritzmg:
Closing, as I don't think anything needs to be changed here.

PR #388, opened 2020-05-25
Fix description

Tastaturberuf:
Use same as here: https://docs.contao.org/dev/reference/hooks/loadFormField/

### DIFF
--- a/docs/dev/reference/hooks/validateFormField.md
+++ b/docs/dev/reference/hooks/validateFormField.md
@@ -19,9 +19,9 @@ object as return value.
 
     Object of the current front end widget. Use it to access form field properties.
 
-2. *int* `$intId`
+2. *string* `$formId`
 
-    ID of the `tl_form_field` record.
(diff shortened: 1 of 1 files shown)


docs/dev/reference/hooks/validateFormField.md
```diff
@@ -19,9 +19,9 @@ object as return value.
 
     Object of the current front end widget. Use it to access form field properties.
 
-2. *int* `$intId`
+2. *string* `$formId`
 
-    ID of the `tl_form_field` record.
+    Alias of the current form. Used in the `value` attribute of the hidden form field `FORM_SUBMIT`. Don’t confuse with `$form->id`.
```
fritzmg:
```suggestion
    Alias of the current form. Used in the `value` attribute of the hidden form field `FORM_SUBMIT`. Do not confuse with `$form->id`.
```

fritzmg (changes requested):

fritzmg (approved):

PR #410, opened 2020-06-04
Update callbacks.md

Ernestopheles:
good to know its not only BE involved


fritzmg:
There are a few more callbacks that can be executed in the front end, not just this one. I'd rather have mentioned them all somehow.

fritzmg:
I have added a notice in 02ffd86

PR #425, opened 2020-06-08
Update frontend-modules.de.md

Ernestopheles:
https://contao.slack.com/archives/CK4J0KNDB/p1591364869019400


fritzmg (changes requested):
This has nothing to do with the newsletter reader module itself. It is generally the case for all "reader" modules. 

Also, usually users would not consciously integrate two reader modules on the same page. The problem you describe happens, if you erroneously configure an event- or newslist module to show an event-/newsreader module automatically. So instead, that feature should be documented more clearly, in order to avoid that pitfall.

fritzmg:
Closing in favor of https://github.com/contao/docs/pull/424

PR #428, opened 2020-06-10
Document the issue workflow on GitHub

leofeyer:


Toflar (commented):
I would write it as if you were explaining it to a reader that's not familiar with our internals. Right now you are writing it as if you were explaining it to a core dev. But imho it should be documented for the world to read :)
I'd just use passive style writing. E.g. instead of "After reading a new issue, please label it either as" just use "New issues either get the bug or feature label assigned, depending on the nature of the issue"...

Content-wise, it's all good 👍

leofeyer:
> I'd just use passive style writing.

Hm, active voice is generally considered better than passive voice. 

> Right now you are writing it as if you were explaining it to a core dev.

Well, that is because I am. 😄 Who else would be interested?

Toflar:
Because everybody is interested in the meaning of the labels?

leofeyer:
@fritzmg What do you think?

fritzmg:
I agree with @Toflar that everybody will be interested in what the labels mean and how the process is done. But of course it should also be directed at the developers. 🤔

fritzmg (approved):
With regards to content I think this is fine. We can always change the phrasing later on, if we need or want to.

PR #431, opened 2020-06-11
Add and use new Shortcode taxonomylist

fkaminski:
New Shortcode: taxonomylist:
Create a List with linked Pages matching one or more Taxonomy Terms (similar to the shortcode "children"). 

The Taxonomic Context could be "categories" or "tags".
(Here with Front Matter "tags:" - as discussed with @fritzmg on slack)

### DIFF
--- a/docs/manual/guides/add-first-index-page.de.md
+++ b/docs/manual/guides/add-first-index-page.de.md
@@ -6,6 +6,10 @@ aliases:
     - /de/anleitungen/die-erste-startseite/
 weight: 10
 hidden: true
+tags: 
+   - "Theme"
+   - "Seitenlayout"
+   - "Seitenstruktur"
 ---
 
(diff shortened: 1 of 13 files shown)


fritzmg:
Looks great, I'll have a look at it in more detail, if I see any issues :)

fkaminski:
As already mentioned in slack: Finding the right tag term is difficult. 
Please check this regardless of the shortcode ...

fritzmg:
The shortcode looks good. I made a small change to fix the invalid HTML.

fkaminski:
Thank you

fkaminski:
I wanted to sum up guides (like grid, SCSS, CSS, fonts etc.) with the tags term "Layout". But at the end its "Theme" - right? Same for "Seitenlayout". To reduce the number of tags i would like to use only these: 

Theme, Modul, Seitenstruktur, Template

What do you think?

netzarbeiter:
I like

Toflar:
Nice!

PR #432, opened 2020-06-11
Add clarification about readers and the reader setting in list modules

fritzmg:
This is something that people in the community get frequently wrong. They select an eventreader in the eventlist module (or in recent Contao versions also a newsreader in the newslist module), mostly without really knowing what kind of effect this actually has. Doing so might incapacitate your reader pages, if  the list module is placed in the page layout (and thus on every page).

This PR adds some clarification about that and also mentions that you can't have more than one reader module on the same page.


netzarbeiter:
Thank you, Fritz.

PR #436, opened 2020-06-13
Show how to remove a field from a subpalette

georgpreissl:
It should be shown how to remove a field from a subpalette.
Since there is no second argument needed as when removing a field from a regular palette.


fritzmg (approved):

fritzmg:
Thank you @georgpreissl :)

PR #441, opened 2020-06-15
Typo: Form instead of From

rorych:
I think this is a typo, where getPageIdFormUrl was used instead of getPageIdFromUrl.


fritzmg:
Yes, thank you :). I have reverted your change to the aliases and added the new alias. The point of aliases is to have redirects for the listed URL aliases. This way the old URL will still work.

PR #447, opened 2020-06-19
Update frontend-modules.de.md

Ernestopheles:
Einige Ergänzungen


fritzmg:
I don't think we should add documentation for other extensions there.

Ernestopheles:
So any idea where to add this documentation?

Am Freitag, den 19.06.2020, 11:13 -0700 schrieb Fritz Michael
Gschwantner:

PR #455, opened 2020-06-29
Enable framework.assets by default

bytehead:
See https://github.com/contao/contao/pull/1883


fritzmg:
Should we add a line break after that? Or may be remove all the empty lines within `framework:`? What do you think?

bytehead:
I'm fine with removing the empty lines. 👍

fritzmg (approved):

PR #461, opened 2020-07-04
Do not use blacklist and whitelist

fritzmg:
Docs PR for https://github.com/contao/contao/pull/1798

Btw. @Toflar shouldn't we also document the `COOKIE_REMOVE_FROM_DENY_LIST`/`COOKIE_DISABLE_FROM_BLACKLIST` env vars?

### DIFF
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -213,7 +213,11 @@ That's why, after changing this value, you should regenerate the application cac
 users. For more information please visit the [Symfony documentation](https://symfony.com/doc/current/reference/configuration/framework.html#secret).
 
 
-### `COOKIE_WHITELIST`
+### `COOKIE_ALLOW_LIST`
+
+{{% notice info %}}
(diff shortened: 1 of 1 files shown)


Toflar (changes requested):
Maybe we should write in 4.8 and 4.9? Or don't we care because unmaintained version?

> Btw. @Toflar shouldn't we also document the COOKIE_REMOVE_FROM_DENY_LIST/COOKIE_DISABLE_FROM_BLACKLIST env vars?

Only the first one, it's a new feature in 4.10 only.

fritzmg:
> Or don't we care because unmaintained version?

We don't care, because unmaintained, yes.

> Only the first one, it's a new feature in 4.10 only.

Ah ok, did not realise. Then a ticket/PR for the new feature is missing, can you create one?

Toflar:
Done in https://github.com/contao/docs/issues/463

Toflar:
You can turn this into a regular PR then, because the statement is correct.

PR #473, opened 2020-07-20
Use correct template path

Toflar:
The root `/templates` path is meant to be used to override existing templates. If you write app templates, put them into `contao/templates`.

### DIFF
--- a/docs/dev/framework/content-elements.md
+++ b/docs/dev/framework/content-elements.md
@@ -99,7 +99,7 @@ Using the naming convention for templates mentioned above, the final template na
 for this content element will be `ce_my_content_element`:
 
 ```html
-<!-- templates/ce_my_content_element.html5 -->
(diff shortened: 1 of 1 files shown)


fritzmg:
Hm.. is that distinction really necessary? Because then technically templates like `ce_text_foo` etc. also need to go into the `contao/templates/` folder, since those are unique templates and not overriding templates.

Toflar:
Everything you want people to be able to override in the back end should go to `contao/templates`. I think it makes a difference because I don't want people to edit things in `/templates` :)

fritzmg:
True. But then we should document this more clearly, e.g. in the Getting Started and Templates section.

fritzmg:
> But then we should document this more clearly, e.g. in the Getting Started and Templates section.

I've made a separte ticket for that » https://github.com/contao/docs/issues/552

PR #475, opened 2020-07-21
Bugfix?

fiedsch:
I just saw this 

> Die beschriebenen Tastaturkürzel funktionieren in dieser Form nur unter Windows. Mac-Nutzer müssen anstatt der [Alt]-Taste [Ctrl]+[⌥ Opt] in Verbindung mit dem jeweiligen Kürzel verwenden. 

in the manual. The ` Opt` is not part of the keyboard-shortcut, is it?

See PR if it needs to be removed.

I did not find other occurences of `⌥` in the docs, so this should be the only place affected.


Toflar:
For Mac, this key is not called "Alt(ernative)" but "Opt(ion)" instead :) So it's actually correct or it adds at least correct information on top of the sign :) 
See e.g. https://support.apple.com/en-us/HT201255

fiedsch:
OK, that makes sense. My MacBook's keyboard and my "big" keyboard only show ⌥ (together with alt), which is, why I was confused. In addition the `Opt` looks a bit like `0pt` which confused me even more.

Issue #484, opened 2020-07-22
Change name of variable in example of picture factory


zonky2:
https://docs.contao.org/dev/framework/image-processing/image-picture-factory/#example

the first and second example has as varaible `$image` - the third example `$resizedImage` ...

To avoid misunderstandings it might be better to name the variable in all three examples the same as `$image` or `$resizedImage`

fritzmg:
In the top right corner of the documentation is a link called _Edit this page_, with which you can quickly make a PR for such small fixes and suggestions :)

zonky2:
For changes where I am quite sure, I do this as you know - here I was not sure if there is a special reason...

https://github.com/contao/docs/pull/485

fritzmg:
> here I was not sure if there is a special reason...

We'll find out in the PR ;). Closing in favor of #485

PR #485, opened 2020-07-22
Set variables to the same name

zonky2:
see #484


fritzmg:
@ausi should be fine, right? Or was there a reason behind the different variable names? (Or did that only happen after the split done by @m-vo ?)

ausi (approved):
Yes, should be fine IMO.

fritzmg:
Thank you @zonky2 :)

PR #490, opened 2020-07-25
Add deepl command

fritzmg:
This adds a `deepl` command for translating markdown files within `docs/manual/`. @netzarbeiter with this I will provide a separate branch with all current articles translated by DeepL.

### Setup

A `DEEPL_API_KEY` environment variable with your DeepL API key needs to be present before executing the command.

### Usage

Executing `deepl/deepl translate` will automatically translate all German files to English, where the English translation is missing. 

```
$ deepl/deepl translate
Translating administration-area/call-and-structure-of-the-backend.de.md » administration-area/call-and-structure-of-the-backend.en.md
Translating administration-area/edit-data-records.de.md » administration-area/edit-data-records.en.md
Translating administration-area/list-data-records.de.md » administration-area/list-data-records.en.md
Translating article-management/articles.de.md » article-management/articles.en.md
…
Used 10827 characters out of 1000000.
```

The other way around works too:

```
$ deepl/deepl translate en de
Used 10827 characters out of 1000000.
```

You can also translate single files:

```
$ deepl/deepl translate --file layout/templates/template-inheritance.de.md
Translating layout/templates/template-inheritance.de.md » layout/templates/template-inheritance.en.md
Used 10956 characters out of 1000000.
```

### Fixup

The translation will often break some of the yaml structure, code blocks and shortcodes slightly (e.g. missing quotes, backticks and curly braces). The translation command makes no attempt at fixing that and it will always have to be manually adjusted. However, you can use the `fixup` command to at least fix some of the issues:

```
$ deepl/deepl fixup en
Fixing file administration-area/back-end-keyboard-shortcuts.en.md
Fixing file administration-area/_index.en.md
Fixing file extensions/animated-timeline.en.md
…
```
```
$ deepl/deepl deepl:fixup --file installation/system-requirements.en.md
Fixing file installation/system-requirements.en.md
```
This command only fixes the following things though:

* ``` ``php``` » ```` ```php ````
* `{{{%` » `{{%`
* `` `foo'`` » `` `foo` ``

All other issues must be fixed manually.


fritzmg:
As discussed with @netzarbeiter , the current output is not that great because DeepL splits sentences at new lines by default. However, there is a `split_sentences` parameter which can be set to `0` (everything is treated as one sentence), `1` (sentences are split by new lines, which is the default) and `nonewlines` (which splits on interpunction only, ignoring newlines). The latter should improve the quality of the output.

Toflar:
Nice one! Instead of having a fixup command, can't we have deepl ignore certain sections by marking them? Maybe by converting the code to HTML first and then converting back to MD?

fritzmg:
You mean: converting it to HTML, then going through the DOM and send the text of each node to the API sequentially (while ignoring certain nodes) and then rebuild it into markup? I am not sure that going from HTML back to Markdown is all that straight forward, especially since the Markdown is enriched with shortcodes and the like.

Toflar:
Yeah those shortcodes need manual handling anyway, no matter if we convert it to html first or not. But yes, excluding everything that's between the tripple ` etc.

fritzmg:
This has been completely overhauled now. Translating a file processes the following steps now:

1. The meta data header containing YAML data is extracted and processed separately.
    1. The `url` is removed, since that is not needed for the English language, nor should it be translated by the machine.
    2. The `aliases` are set according to the file path and name.
    3. The `title`, `menuTitle` and `description` will be translated.
2. The body of the file is converted to HTML using [`erusev/parsedown`](https://github.com/erusev/parsedown).
3. The nodes of the resulting HTML are processed recursively using [`phpgt/dom`](https://github.com/PhpGt/Dom).
    * `pre` and `code` nodes are skipped.
    * `p` or text nodes are translated.
4. The resulting HTML is converted back to Markdown using [`league/html-to-markdown`](https://github.com/thephpleague/html-to-markdown), using a custom converter for tables (which are not supported out of the box).
5. Hugo shortcodes get broken during translation and are thus getting fixed in the resulting Markdown.
6. A warning is inserted at the top that this is a machine translated file.
7. The previously processed meta data is prepended and then saved to the destination file.

This produces pretty excellent results. The only thing that is missing (at least from my tests during development) are the automatic line breaks after 80 or before 140 characters. But I decided not to invest more time in this, as this isn't that trivial (since you don't want to break up phrases like `**this is bold**`).

```
$ ./deepl translate --help
Description:
  Automatically translates pages from the manual.

Usage:
  translate [options] [--] [<source_lang> [<target_lang>]]

Arguments:
  source_lang           The language to be translated from. [default: "de"]
  target_lang           The language to be translated to. [default: "en"]

Options:
      --file=FILE       A single file to translate. Path must be relative to the docs/manual/ directory.
      --force           Forces a file to be translated, even if the translated file already exists.
```

PR #530, opened 2020-08-10
Document `useRawRequestData`

dmolineus:
This PR adds documentation for the `useRawRequestData` eval flag


fritzmg (approved):

fritzmg:
Thank you @dmolineus :)

Toflar:
I think we should add a warning here. You should make sure the data is never output anywhere in the back end unescaped which it would if you added the field to a regular back end list view for example.

PR #531, opened 2020-08-10
Fix broken table

dmolineus:
This is a follow up PR for #530 which broke the rendered table


fritzmg (approved):

fritzmg:
Right, didn't catch that either.

PR #534, opened 2020-08-13
Feature/image sizes

rabauss:
- Add notice tip for image size using from config file
- Update image size formats example analogous to the options in the backend

### DIFF
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -107,9 +107,9 @@ contao:
 
                     # Examples (for webp settings analogous to the options in the backend):
                     png: [webp, png]
-                    gif: [webp, gif]
                     jpg: [webp, jpg]
                     jpeg: [webp, jpeg]
+                    gif: [webp, gif]
 
                     # Prototype
                     source:               []
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -106,18 +106,10 @@ contao:
                 formats:
 
                     # Examples (for webp settings analogous to the options in the backend):
-                    png:
-                        - webp
-                        - png
-                    jpg:
-                        - webp
-                        - jpg
-                    jpeg:
-                        - webp
-                        - jpeg
-                    gif:
-                        - webp
-                        - gif
+                    png: [webp, png]
+                    gif: [webp, gif]
+                    jpg: [webp, jpg]
+                    jpeg: [webp, jpeg]
 
                     # Prototype
                     source:               []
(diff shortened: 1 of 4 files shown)


docs/manual/system/settings.de.md
```diff
@@ -305,12 +305,19 @@ contao:
...
                     jpg:
                         - webp
                         - jpg
+                    jpeg:
+                        - webp
+                        - jpeg
                     gif:
-                        - png
+                        - webp
+                        - gif
```
ausi: I think the short array syntax would make this section way more readable, WDYT?
```yaml
    png: [webp, png]
    gif: [webp, gif]
    jpg: [webp, jpg]
    jpeg: [webp, jpeg]
```

ausi (approved):

docs/dev/reference/config.md
```diff
@@ -106,18 +106,10 @@ contao:
...
-                    jpeg:
-                        - webp
-                        - jpeg
-                    gif:
-                        - webp
-                        - gif
+                    png: [webp, png]
+                    gif: [webp, gif]
+                    jpg: [webp, jpg]
+                    jpeg: [webp, jpeg]
```
fritzmg, line 111: This will probably be reverted in the future, as this part of the documentation is just the output from `config:dump-reference`.

docs/dev/reference/config.md
```diff
@@ -105,12 +105,11 @@ contao:
...
-                    jpg:
-                        - webp
-                        - jpg
-                    gif:
-                        - png
+                    # Examples (for webp settings analogous to the options in the backend):
+                    png: [webp, png]
+                    jpg: [webp, jpg]
+                    jpeg: [webp, jpeg]
+                    gif: [webp, gif]
```
fritzmg, line 112: This will probably be reverted in the future, as this part of the documentation is just the output from `config:dump-reference`.

rabauss: where can/should we add this as example?

fritzmg: Directly in the configuration of the `contao/core-bundle`. However, within the documentation it would make sense to describe this part separately. imho we should introduce a new article called "Manage image sizes" in https://docs.contao.org/manual/en/layout/theme-manager/, explain the image sizes options of the back end there and then also mention the possibility to configure image sizes in the `config.yml` directly, plus the examples.

A lot of work for sure, but having that would be worth it :) » https://github.com/contao/docs/issues/551

fkaminski: Possibly with links to (or text excerpts from):

- https://blog.kulturbanause.de/2014/09/responsive-images-srcset-sizes-adaptive/
- https://rocksolidthemes.com/de/contao/blog/responsive-images-picture-contao

fritzmg:
@rabauss in the mean time I have added more info to the image processing article. I think this PR can be closed for now :)

PR #537, opened 2020-08-19
Field options callback parameter can be null

koertho:
This small change update the field options callback parameter as it can be null.


fritzmg (approved):

fritzmg:
Thank you @koertho 

Though I think this is the case for more of these callbacks.

PR #587, opened 2020-10-20
Use security voters instead of user instances

aschempp:


fritzmg:
Does this work in Contao 4.4 as well?

Also the downside to doing it like this is that you still have to fetch the user afterwards, whereas before you already had the user.

aschempp:
no this will not work in Contao 4.4, because it does not use Symfony Security.
Why would you fetch the user afterwards?

fritzmg:
Well in most of my use cases I need information about the user, but may be not in all use cases, yeah.

fritzmg:
Thank you @aschempp :)

Issue #478, opened 2020-07-21
Document page controllers and (may be) content composition


fritzmg:
See https://github.com/contao/contao/pull/1516

Page types can now be registered as controllers via the `@Page` annotation for example, where you can define largely the same settings as with regular routes, plus some additional ones (e.g. relative or absolute routes).

The aforementioned PR also describes a new concept of content composition, which we can may be cover here too.

I'll create a PR for @aschempp to review.

fritzmg:
See #604

PR #604, opened 2020-11-01
Add page controllers and routing info

fritzmg:
Implements #478 

This adds documentation about the new page controllers and also adds more information about routing in general, i.e. Contao specific route attributes and how and why to use them and also a general guide on how to implement your own independent routes.

@aschempp As far as I know there is nothing place yet to make rendering a complete layout defined via the page layout of Contao for a custom page type implemented via page controllers, is that correct? You can enable `contentComposition` and then are able add articles and content elements in the back end for that page - but you would still have to render everything yourself.

@netzarbeiter I've marked this for the two month review :)


docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+
+Starting with Contao **4.10** you can implement so called _Page Controllers_. These
+are special page types implemented as controllers in order to handle the request
+to the route of a specific page type within the Contao site structure.
+
+
+## Registering Page Controllers
+
+As with content elements, front end modules, hooks and DCA callbacks, Page controllers
+can also be registered via annotations. The following shows the most basic example:
```
aschempp:
```suggestion
can be registered via annotations. The following shows the most basic example:
```

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+{
+    public function __invoke(): Response
+    {
+        return new Response('Hello World!');
+    }
+}
+```
+
+The same can be achieved without annotations by tagging the respective service with 
+`contao.tag`.
```
aschempp:
```suggestion
`contao.page`.
```

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+A translation for the back end label should be defined to:
+
+```php
+// contao/languages/en/default.php
+$GLOBALS['TL_LANG']['PTY']['example'] = ['Example', 'Example page type.'];
+```
+
+Now we are all set and can add this new page type in the site structure of the Contao
```
aschempp:
```suggestion
Now we are all set and can add this new page in the site structure of the Contao
```

please refrain from calling it `page type`, thats the _old term_. There are different types of pages, and there are different types of content elements or modules. But we never call them `content element type`.

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+back end:
+
+![Custom page type in the Contao back end](/framework/images/custom-page-type-back-end.png?classes=shadow)
```
aschempp, line 73: I don't see that image in the pull request?

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+you can pass it as the first parameter of the annotation (or use `type="custom_type"`).
+
+```php
+/**
+ * @Page("custom_type")
+ */
+```
+
+Note that this one of the differences between the `@Page` and Symfony's `@Route`
+annotation where in the latter case, the first parameter is the _path_ of the route.
```
aschempp, line 137: 👍 👍

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+For instance, with the following annotation:
+
+```php
+/**
+ * @Page(path="/foo/bar")
+ */
+```
+
+the URL of the page will _always_ be `https://example.com/foo/bar`, no matter what
+is defined in the back end.
```
aschempp: add something like 

> => means the `alias` field should not be in the palette

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+So for example, with the following annotation:
+
+```php
+/**
+ * @Page(path="foo/bar")
+ */
+```
+
+and an alias like `example/alias` defined in the back end, the final front end URL
+of the page will be `https://example.com/example/alias/foo/bar`.
```
aschempp: what about the URL suffix?

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+that URL suffix in the page controller's configuration:
+
+```php
+/**
+ * @Page(urlSuffix=".csv")
+ */
+```
+
+So if the page in the site structure has the alias `foo/bar` then the final front
+end URL will be `https://example.com/foo/bar.csv`.
```
aschempp: > … even though the root page's url suffix might be `.html`

docs/dev/framework/page-controllers.md
```diff
@@ -0,0 +1,190 @@
...
+
+
+### `contentComposition`
+
+This is a boolean property defining whether this page type is used for 
+_content composition_. By default, custom page types implemented via page controllers
+do not have content composition enabled, meaning you are not able to add page articles
+to that page in the back end. If your page handles the rendering of page articles
+then you need to enable this property on your page controller:
```
aschempp: it's not specifically about articles, it is related to page layouts. _Content composition_ refers to layouting the page in the back end (including articles). For example, an RSS feed page type does not use content composition.

docs/dev/framework/routing.md
```diff
@@ -0,0 +1,355 @@
...
+    }
+}
+```
+{{% /notice %}}
+
+
+## Request Attributes
+
+When defining a route Symfony allows you to set some default parameters for the
+request handled by this route. As of Contao **4.9** there are two different special
```
aschempp: which ones are for Contao 4.9?

aschempp (changes requested):
> @aschempp As far as I know there is nothing place yet to make rendering a complete layout defined via the page layout of Contao for a custom page type implemented via page controllers, is that correct? You can enable contentComposition and then are able add articles and content elements in the back end for that page - but you would still have to render everything yourself.

You don't render them yourself. You pass the `pageModel` to `PageRegular::renderPage`. However, this is currently "dangerous" because permissions will not be checked. See https://github.com/contao/contao/pull/2015/files#diff-84c54a7b63df1afc1990d4aea951a7a696d445b16a2c4937591831a96f5d1ea4R28 for an early draft.

fritzmg:
> You don't render them yourself. You pass the `pageModel` to `PageRegular::renderPage`. However, this is currently "dangerous" because permissions will not be checked. See https://github.com/contao/contao/pull/2015/files#diff-84c54a7b63df1afc1990d4aea951a7a696d445b16a2c4937591831a96f5d1ea4R28 for an early draft.

Hmm... should we even mention it then? Still seems rather involved.

aschempp:
mention page rendering or the page controllers at all?

docs/dev/framework/page-controllers.md
fritzmg: Yeah, forgot to create it :)

fritzmg:
> mention page rendering or the page controllers at all?

Mentioning how you would be able to render a page layout in your own page controller. Explaining everything that would be in the `AbstractPageController` from your PR would be too much.

docs/dev/framework/page-controllers.md
fritzmg: Oh actually it is in the PR.

docs/dev/framework/page-controllers.md
fritzmg: Ok, rephrased now.

docs/dev/framework/routing.md
fritzmg: Not sure why I phrased it this way. Both of them are available in Contao 4.4.

aschempp:
I think it really depends on the use case. I implemented similar controller stuff years ago in Contao <4.9, but simply because I knew my system does not need authentication or thelike. If the (page) controller is created for the app, I might control what's necessary. But it shouldn't be in a bundle just yet because people might expect certain behavior 🤷

fritzmg:
Ok, so the most basic example would be

```php
// src/Controller/Page/ExamplePageController.php
namespace App\Controller\Page;

use Contao\CoreBundle\ServiceAnnotation\Page;
use Contao\PageModel;
use Contao\PageRegular;
use Symfony\Component\HttpFoundation\Response;

/**
 * @Page(contentComposition=true)
 */
class ExamplePageController
{
    public function __invoke(PageModel $pageModel): Response
    {
        // The legacy framework relies on the global $objPage variable
        global $objPage;
        $objPage = $pageModel;

        // Render the page using the PageRegular handler from the legacy framework
        return (new PageRegular())->getResponse($pageModel, true);
    }
}
```
However, doing just that is pretty nonsensical as you could simply use a regular page instead. So I'd like to show some use cases, but I am not actually sure what the use cases are. Can you provide some simplified real world examples?

aschempp:
In the current state (where there's no abstraction to render a layout), custom page controllers mostly make sense if you want to generate something else than a page. As an example, the news extension allows to define RSS feeds. Instead of the current configuration and writing the feed into a folder, the feed could be a page controller. For each feed, you would create a page in the tree and give it the alias of the feed path you want to have. The configuration of the feed would take place inside the page configuration. This would have the advantage of a feed being specific to a domain/root page, and you could use any page alias/folder alias you want to reach the news feed (even the same as the list page, with a different suffix).

fritzmg:
> However, this is currently "dangerous" because permissions will not be checked.

Btw. that applies to _all_ pages implemented as page controllers, not just the ones using content composition, does it not? If you implement a page via a page controller and then allow it to be restricted to a front end member group for example, you still need to implement that check yourself.

docs/dev/framework/page-controllers.md
fritzmg: That's hard to do btw. There even is a `FilterPageTypeEvent` in the core now ;)

fritzmg:
I have added some additional notes on how or why to use page controllers. I also added the primitive example for pages with enabled content composition. This will be then later on changed for the upcoming Contao versions.

PR #612, opened 2020-11-14
Add services.yaml for migration example

eknoes:
This adds also the content of the services.yaml file for the contao:migrate developer example

### DIFF
--- a/docs/dev/framework/migrations.md
+++ b/docs/dev/framework/migrations.md
@@ -108,6 +108,17 @@ class CustomerNameMigration extends AbstractMigration
 }
 ```
 
+The corresponding `services.yaml` looks like this:
+```yaml
+# config/services.yaml
+services:
+    App\Migration\CustomerNameMigration:
+        arguments:
+            - '@database_connection'
+        tags:
+            - { name: contao.migration, priority: 0 }
+```
+
 
 ## Read more
 


fritzmg:
That should not be necessary, if you have `autoconfigure` enabled (which it is by default in Contao 4.9). Did you have any issues?

fritzmg:
Just to clarify what I mean: by default in the Contao Managed Edition since Contao 4.9, without having your own `services.yaml`, all classes within the `App\` namespace will automatically be registered as a service, with `autowire` and `autoconfigure` enabled.

Thus, if your migration class either extends from `AbstractMigration`, which implements `MigrationInterface` _or_ your migration class implements `MigrationInterface` directly, the service will automatically be tagged with `contao.migration` due to autoconfiguration (due to the implementation of the `MigrationInterface`).

If you _do_ use your own `config/services.yaml` then the same applies, as long as you enabled `autoconfigure` in it, as suggested in the documentation: https://docs.contao.org/dev/getting-started/starting-development/#autoloading-services-and-actions

```yml
# config/services.yml
services:
    _defaults:
        autowire: true
        autoconfigure: true
        …
```

If you do not use autoconfiguration, then the usage of the `contao.migration` tag is already explained in the documentation here: https://docs.contao.org/dev/framework/migrations/#definition

```yml
# config/services.yaml
services:
    App\Migration\MyMigration:
        tags:
            - { name: contao.migration, priority: 0 }
```

eknoes:
Thanks for the clarification, it might not be necessary then! I am not that confident with Symfony and Contao4 yet, I just had to migrate a small extension to the composer structure from the "old" Contao structure and it didn't work, thus I guess I do not have autowiring enabled yet. That led to an unusable installer and an unusable `contao:migrate` tool which both threw an exception.

PR #619, opened 2020-11-20
Add answer if CM has hung up

zonky2:
Add answer if CM has hung up


fkaminski:
Maybe this hint would also be something for the FAQ:
https://docs.contao.org/manual/de/faq/#contao-manager

zonky2:
what's the next step here?

netzarbeiter:
I had to translate it and add it to the FAQ. PR can only be merged with the translation.

PR #620, opened 2020-11-22
Use markdown code blocks instead of HTML

fritzmg:
Instead of using HTML I think we should just use standard markdown code blocks here, like it is done in the rest of the documentation.

Screenshot_2020-11-22 Über den Contao Manager Contao Handbuch

### DIFF
--- a/docs/manual/installation/contao-manager.de.md
+++ b/docs/manual/installation/contao-manager.de.md
@@ -178,8 +178,11 @@ weiteres Benutzerkonto hinzufügen. In unserem Fall ist das <code>h.lewis</code>
 {{% notice info %}}
 Der Wert für »password« muss hierbei verschlüsselt eingetragen werden. Über [bcrypt-generator.com](https://bcrypt-generator.com/)
 könntest du z. B. den notwendigen Hash-Wert generieren. Alternativ dazu, kann man den Hash-Wert auch mit dem folgenden Konsolenaufruf
-in seiner eigenen Contao-Installation erstellen:<br>
-<code>php vendor/bin/contao-console security:encode-password 'my_1._pA~~w0rd'</code>
+in seiner eigenen Contao-Installation erstellen:
+
+```bash
+php vendor/bin/contao-console security:encode-password 'my_1._pA~~w0rd'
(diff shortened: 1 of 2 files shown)


netzarbeiter:
Standard markdown is not working in `notices`.

fritzmg:
No, standard markdown does work. Only certain things do not work (like link references), and you can have only one paragraph (due to the theme's CSS).

The problem with the original version of the markdown was the missing line breaks.

netzarbeiter:
Doesn't worked here. https://github.com/contao/docs/commit/bf1e9bd870c8712585e7942d2e50873e4e817505

fritzmg:
Yes, because of the wrong syntax. As I said, there are line breaks missing before and after the three back ticks.

netzarbeiter:
But line breaks with two spaces are not working.

fritzmg:
> But line breaks with two spaces are not working.

What do you mean with two spaces? The code from this PR is working (see screenshot). It is used exactly like this in many places in the documentation. Examples:

* https://docs.contao.org/dev/framework/cron/#command-line
* https://docs.contao.org/dev/framework/manager-plugin/#the-package-specific-manager-plugin
* https://docs.contao.org/dev/getting-started/extension/#service-configuration
* https://docs.contao.org/dev/getting-started/starting-development/#application-configuration
* https://docs.contao.org/manual/de/system/einstellungen/#website-suche
* https://docs.contao.org/manual/de/system/einstellungen/#e-mail-versand-konfiguration
* https://docs.contao.org/manual/de/system/einstellungen/#verschiedene-e-mail-konfigurationen-und-absenderadressen
* …

Just to clarify: this is what @zonky2 used in his original PR:

````
```php vendor/bin/contao-console security:encode-password 'my_1._pA~~w0rd'```
````

which is incorrect markdown syntax (_note:_ this does work in Slack, may be that is why he wrote it this way).

The correct way is the following:

````
```
php vendor/bin/contao-console security:encode-password 'my_1._pA~~w0rd'
```
````

In my PR I additionally defined

````
```bash
````

to signify that it is a shell command.

PR #621, opened 2020-11-22
[RFC] Add version label

fritzmg:
Currently the article explaining all the page settings is a little difficult to read, because there are so many options that have been introduced in later Contao versions. I propose to introduce a new shortcode `{{< version-tag "..." >}}` which is simply an inline version of the other shortcode, so that you can use it more easily in lists.

For example, this is the page before the change:

before

And this is it after (included in this PR):

after

@netzarbeiter what do you think?


netzarbeiter:
I think the idea is very good. Thank you, Fritz.

PR #629, opened 2020-12-07
Add info about page model in fragments and page controllers

fritzmg:
See https://github.com/contao/contao/pull/2506

I also added some general info about being able to pass the current page's model into your controller action as an argument. @aschempp I only added this to the page controller documentation, I am not sure if there is a use case outside of page controllers as well, apart from fragments?


aschempp:
Sorry to say, but I still think this should not be added to the documentation. This would mean we promote it as a good practice to get the page object from the request. The reason it is there is for routing, not for users in a controller to retrieve it. Either use the controller argument resolver, or the ("officially supported") method added in Contao 4.9.10. For any other case, I would prefer people sticking to the "old habits" of `global $objPage`, these are the two "APIs" we support. There's no guarantee the request attribute will exist in the future.

fritzmg:
> but I _still_ think this should not be added to the documentation

You haven't expressed this before 😁. I am fine with removing it, I assumed it was best practice to use it and past slack conversations seemed to indicate that too.

fritzmg:
But that is unrelated to this PR anyway. This PR only adds the argument.

aschempp:
hmm, you mean because 

> You could then use the request object to retrieve the page model of the current 
 page via the [`pageModel` request attribute][PageModelRequestAttribute].

means it's already in the documentation ?

fritzmg:
Yes, it's already in the documentation, see https://docs.contao.org/dev/framework/routing/#page-model

aschempp:
well… can we remove it? 😇 or maybe make it up for discussion for the next public call?

fritzmg:
Ok, I made a separate ticket for that. What about this PR though?

aschempp:
It seems technically correct (even though not very explanatory, but maybe it does not need to be). Feel free to merge it if you think so, we would just (eventually) remove it again with https://github.com/contao/docs/issues/631 right?

fritzmg:
> (even though not very explanatory, but maybe it does not need to be)

What could be improved?

> Feel free to merge it if you think so, we would just (eventually) remove it again with #631 right?

Well no, just the part about the request attribute would be removed. It is still best practise to get the model of the current page via argument resolving, is it not?

aschempp:
I think the model argument resolving is hard to understand by itself. The following part

> In Symfony you can require the current `Request` object to be passed into your invokable
 controller or action method as an argument

can be confusing in my opinion. It basically means that Symfony does not only support dependency injection through constructor or method call, but also on controller arguments. And there are a million possibilities for that, you can get any container service (afaik) or any request attribute from it. They are tickets to add query parameter support (afair through PHP8 annotations).

I think it depends on whether you want to explain how Symfony controller argument resolving works (then it's too little information, and the link to Symfony only explains how to add your own), or we target developers of front end modules or content elements, where the request is already present in the abstract controller, so there's not need to "require the current Request". 🤷

fritzmg:
> or we target developers of front end modules or content elements

The PR is also targetted to developers of page controllers.

aschempp:
I see. So we can basically tell them they can do whatever Symfony developers can?

fritzmg:
Well, yeah, is that not the point? ;)

PR #659, opened 2021-02-13
Update callbacks, missing select.buttons

danielkochengineer:
I found out that the select.buttons is missing in the documentation so I tried to add it. I hope I did all correct since this is my first change to a Contao documentation, but I looked at the general rules to make sure it is.


fritzmg (approved):
Good catch! I'll may be merge this subsection later on with the previous one, since they are functionally similar - just two different use cases (one is for the buttons when editing/creating a record, the other one for copying/overriding/editing records when selecting multiple ones).

fritzmg:
Thank you @danielkochdakitec

PR #660, opened 2021-02-14
The type of the 'type' tag should be 'string'

markocupic:


fritzmg (approved):

fritzmg:
Thank you @markocupic !

PR #664, opened 2021-02-21
Clarify back end keyboard shortcuts

fritzmg:
Small clarification for the back end keyboard shortcuts, including a link to developer.mozilla.org where it is explained in more detail, which key combination needs to be used under which operating system and browser.


netzarbeiter (approved):
Thank you, Fritz

PR #666, opened 2021-02-21
Link to Devilbox documentation for Docker and Docker Compose installation

fritzmg:
This is an alternative PR for https://github.com/contao/docs/pull/663. Instead of also explaining how to install _Docker_ and _Docker Compose_, which is just a basic requirement for running the Devilbox, we should simply link to the documentation on Devilbox on how to do that, should anyone require any help on how to install those software applications.


netzarbeiter (approved):
Thank you, Fritz

PR #672, opened 2021-02-25
Fix preview script

bytehead:
According to https://github.com/contao/contao/blob/4.9/manager-bundle/src/Resources/skeleton/web/preview.php


bytehead:
🙏

PR #677, opened 2021-03-04
Note that the bundle folder must be in the Contao vendor directory.

georgpreissl:
The path shown so far leads to an error:
https://github.com/contao/core-bundle/issues/1114


fritzmg:
The reason why I haven't merged this yet is because it's not that simple. If you start your development directly in the `vendor/` directory, then you should create a path that is equal to your package name, so that Composer can detect that the source is already present.

Other than that I still want to fix this in the core 🙃

fritzmg:
It seems that this issue does not occur any more, so I am closing this for now.

PR #678, opened 2021-03-05
Add non-HTTPS CSRF cookie to example allow list

ausi:
Local development via unsecure HTTP is still quite common I think. 
And if the `csrf_contao_csrf_token` is missing in the allow list, debugging the CSRF issue is quite bad (I learned the hard way 😅).


fritzmg (approved):

fritzmg:
Thank you @ausi :)

PR #681, opened 2021-03-07
Content composition is enabled by default

richardhj:
The _content composition_ is enabled by default. See https://github.com/contao/contao/blob/6cdb7fce22c393ba19e0ca3b9237398b9bbba050/core-bundle/src/DependencyInjection/Compiler/RegisterPagesPass.php#L61

This PR adjusts the description accordingly.


fritzmg (approved):

fritzmg:
Thank you @richardhj :)

PR #684, opened 2021-03-08
Dieser Artikel ist zu technisch und für normale Benutzer nicht relevant

georgpreissl:
Ich denke, dieser Artikel über Caching ist viel zu detailliert. In dieser Form solte er im Entwicklerhandbuch enthalten sein. Den "normalen" Anwender interessiert wahrscheinlich eher ober er nun bei den Cache-Einstellungen die Checkbox "Cachezeiten festlegen" aktivieren soll oder nicht. D.h. was diese Checkbox bewirkt.

### DIFF
--- a/docs/manual/performance/http-caching.de.md
+++ b/docs/manual/performance/http-caching.de.md
@@ -1,6 +1,6 @@
 ---
 title: "HTTP Caching"
-description: "HTTP Caching mit Contao."
+description: "Technische Aspekte des HTTP Caching mit Contao."
 url: "performance/http-caching"
 aliases:
     - /de/performance/http-caching/


fritzmg:
> Den "normalen" Anwender interessiert wahrscheinlich eher ober er nun bei den Cache-Einstellungen die Checkbox "Cachezeiten festlegen" aktivieren soll oder nicht. D.h. was diese Checkbox bewirkt.

If you want to know what these settings do, then you need to know how the cache works. If this is too technical for someone, then they probably should not use these settings - or hire someone else to develop their site for them.

georgpreissl:
I think it would make more sense to explain in simple terms what "Set Cache Times" does. Because the manual is not aimed at developers.

The normal Contao user is interested in the following, I assume:
- When does it make sense to activate "Set cache times" and why should I do that?
- Are changes to the page immediately visible to a visitor when I activate the cache?
- What is the difference between private and shared cache?

It is possible that the answers to these two questions are hidden somewhere in the article. Very few readers are likely to want to look for it. And I have to admit I haven't found them all yet.

I think this article is simply not appropriate for the target audience.
Be that as it may, it was just a suggestion for improvement on my part.

fritzmg:
> I think it would make more sense to explain in simple terms what "Set Cache Times" does.

That's what https://docs.contao.org/manual/en/layout/site-structure/configure-pages/#cache-settings is for. The article you are referring to is an addition to the documentation in order to explain how HTTP caching in Contao works in more detail.

fritzmg:
> * When does it make sense to activate "Set cache times" and why should I do that?

See the FAQ entry here for example: https://docs.contao.org/manual/en/performance/http-caching/#faq

fritzmg:
> Be that as it may, it was just a suggestion for improvement on my part.

I am sorry to be a bit blunt here, but changing the title of that particular article is not really an improvement and would not change anything on the points you are making. You can always provide PRs for changing the actual content of course, but I do not think there is anything that needs to be changed. Pretty much all your questions are answered in the article and the article is written in a detailed and easy to understand way in my opinion.

netzarbeiter:
ich persönlich finde den Text nicht zu technisch (Caching ist ein sehr technisches und komplexes Thema) und schliesse deshalb das Ticket.

PR #689, opened 2021-03-16
Open search

dmolineus:
This pull request allows to trigger the search by passing a search param. Furthermore it adds open search information for easy installation in supported browsers.

### DIFF
--- a/docs/manual/search.xml
+++ b/docs/manual/search.xml
@@ -3,6 +3,6 @@
     <ShortName>Contao Manual</ShortName>
     <Description>Contao Manual</Description>
     <InputEncoding>UTF-8</InputEncoding>
-    <Image width="16" height="16" type="image/x-icon">[https://contao.org/favicon.ico]</Image>
+    <Image width="16" height="16" type="image/x-icon">https://contao.org/favicon.ico</Image>
     <Url type="text/html" method="get" template="https://docs.contao.org/manual/?search={searchTerms}"/>
 </OpenSearchDescription>
--- a/docs/manual/search.xml
+++ b/docs/manual/search.xml
@@ -1,0 +1,8 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
+                       xmlns:moz="http://www.mozilla.org/2006/browser/search/">
+    <ShortName>Contao Manual</ShortName>
+    <Description>Contao Manual</Description>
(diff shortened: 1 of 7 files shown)


fritzmg (changes requested):
Very nice, thank you!

One change request though: please move

```html
<link rel="search" type="application/opensearchdescription+xml" …>
```

to `page/layouts/partials/custom-header.html` as this is the normal place for any such additions to the `<head>`. The docs currently override the complete `header.html` of the theme only because of [this change](https://github.com/contao/docs/commit/363b812d845a2827599fe236e10e868ed0e01d75) and I keep this file updated with the theme manually at the moment.

dmolineus:
I moved the line to the `custom-header.html` file.

docs/manual/search.xml
```diff
@@ -0,0 +1,8 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/"
+                       xmlns:moz="http://www.mozilla.org/2006/browser/search/">
+    <ShortName>Contao Manual</ShortName>
+    <Description>Contao Manual</Description>
+    <InputEncoding>UTF-8</InputEncoding>
+    <Image width="16" height="16" type="image/x-icon">https://contao.org/favicon.ico</Image>
+    <Url type="text/html" method="get" template="https://docs.contao.org/manual/?search={searchTerms}"/>
```
fritzmg: Will this work? The manual is separated into `en/` and `de/`. Currently `https://docs.contao.org/manual/` just redirects to `https://docs.contao.org/manual/de` via HTML (will be changed in the future).

dmolineus: No, then it won't work. Any idea how to solve it? It's possible to keep the param when doing the redirect or do we need to ship different search.xml files.

fritzmg: Yes, we'll need different `search.xml` files. Hm, what happens if you create a `docs/manual/de/search.xml`, will it be copied over when you run Hugo?

dmolineus: I figured out how to solve it. I use a custom output format for the home page now which creates the `search.xml`.

fritzmg (approved):

fritzmg:
lgtm, thank you @dmolineus :)

dmolineus:
Yeah, it works and I can use my browser search bar now to quickly access the docs. Thanks :-)

PR #694, opened 2021-03-23
Update callbacks - missing serp preview eval.url, eval.title_tag

rabauss:
Currently the field callbacks of the serp preview are not in the docs.
But we have to wait for a fix of the core, because we cannot tag them in a service for now!

### DIFF
--- a/docs/dev/reference/dca/callbacks.md
+++ b/docs/dev/reference/dca/callbacks.md
@@ -558,6 +558,28 @@ a button for an import "wizard".
 {{% /expand %}}
 
 
+### `fields.<FIELD>.eval.url`
+
+Allows you to add an url to the serp preview field.
+
+{{% expand "Parameters" %}}
+* `\Contao\Model` Model object (class from the table)
+
+**return:** `string` URL for the serp preview
+{{% /expand %}}
+
+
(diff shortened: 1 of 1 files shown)


fritzmg:
Can you also provide examples may be? :) Example for an example: 

```md

### `edit.buttons`

Allows you to modify the action buttons at the bottom of record editing form. This
can be used to add additional buttons or remove any of the existing buttons.

{{% expand "Parameters" %}}
* `array` Array of strings
* `\Contao\DataContainer` Data Container object

**return:** `array` Array of strings containing the buttons' markup
{{% /expand %}}

{{% expand "Example" %}}
This example removes the "Save and close" button from the editing form of a record
of the `tl_example` Data Container.

…
{{% /expand %}}
```

rabauss:
Oh nice 👍  - I need to start another PR with examples for other callbacks :-)

fritzmg:
That would be very welcome ;). I wanted to do it every now and again, when I implement a callback that's usable as a general example.

fritzmg (approved):
lgtm!

I'll merge it right away, since it is already merged in Contao and will be available with the next bugfix release.

fritzmg:
Thank you @rabauss

PR #702, opened 2021-03-31
Add hint that panel order matters in DCA (fix #701)

### LINKED ISSUE #701: Document, that panel order in DCAs matter
Hello there,

in a recent project we had the issue that `DC_Table` ignored our `filter`and `search` panels for the `limit` panel and back-end pagination.

You see in the screenshot a search by e-mail-address that results in three entries. The list (not visible here) contains those three entries, but the pagination and the ` panel still take all some hundred entries into account.

The panel configuration in this case was `limit;search;sort`. Because the search panel sets the actual search conditions and values, the limit menu rendered as no search took place. The limit menu then sets the limit for the pagination, hence the pagination is off, too. Defining the panels with `search;sort;limit` fixed this issue.

Conclusion is: The `limit` panel must come **after** `search` and `filter`. 

PR incoming. 

Cheers

bezin:


bezin:
Bug fixed in Contao core 🎉

PR #711, opened 2021-04-24
Human re-write of machine translated

rburch:
Did a human edit for this. The machine translation had:  "third party developers who do not write their own" for the bottom licencing info. I'm pretty sure my change is correct but I don't want say something about the licensing that isn't true.

### DIFF
--- a/docs/manual/introduction/contao-open-source-cms.en.md
+++ b/docs/manual/introduction/contao-open-source-cms.en.md
@@ -6,35 +6,31 @@ aliases:
 weight: 10
 ---
 
-{{% notice warning %}}
-This article is machine translated.
-{{% /notice %}}
-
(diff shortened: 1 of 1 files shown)


fritzmg:
> The machine translation had: "third party developers who do not write their own" for the bottom licencing info. I'm pretty sure my change is correct but I don't want say something about the licensing that isn't true.

I am not even quite sure what the original German source wants to say there 😁. I think @netzarbeiter needs to weigh in here - or someone else who knows where this comes from. Also I think the German source has an error - I don't think "nichteigenständig" exists as a German word.

netzarbeiter:
This Text has been taken from the book "Das offizielle Contao-Buch / Leo Feyer". I have corrected "nicht eigenständig".

netzarbeiter (approved):
Thank you, Bob

PR #717, opened 2021-04-30
dca reference - icon for global operations

patrickjDE:
The global operations section was missing documentation of the ability to add an icon.

Though it's not used by the core afaik, this functionality was eplicitly added (see https://github.com/contao/core/pull/5541)


fritzmg:
Thank you @patrickjDE :)

PR #718, opened 2021-04-30
Update fields.md

Tastaturberuf:
https://github.com/contao/contao/commit/5cf219aecef7f47e45a1c02c1dcb4a786a5cfe0b

### DIFF
--- a/docs/dev/reference/dca/fields.md
+++ b/docs/dev/reference/dca/fields.md
@@ -182,7 +182,7 @@ Each field can be validated against a regular expression.
 | isHexColor         | true/false (`bool`)              | Defines the input as being a color definition in Hex notation. Invalid characters will automatically be removed. |
 | metaFields         | `metaWizard` fields (`array`) | Defines the available fields for the `metaWizard` input type. |
 | custom_rgxp        | Regular expression (`string`) | {{< version "4.11" >}} Custom regular expression to be used when using `'rgxp' => 'custom'` |
-| addWizardClass     | true/false (`bool`) | If false, do **not** add the `wizard` class, if a wizard is present. |
+| addWizardClass     | true/false (`bool`) | Whether or not to add the `wizard` CSS class for the widget if a wizard icon is in use. Default: `true`. _Note:_ this setting does not exist anymore in Contao **4.12** and up. |
 
 {{% notice warning %}}
 Using the `encrypt` option is deprecated and its internal implementation relies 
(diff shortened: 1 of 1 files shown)


docs/dev/reference/dca/fields.md
```diff
@@ -182,6 +182,7 @@ Each field can be validated against a regular expression.
 | isHexColor         | true/false (`bool`)              | Defines the input as being a color definition in Hex notation. Invalid characters will automatically be removed. |
 | metaFields         | `metaWizard` fields (`array`) | Defines the available fields for the `metaWizard` input type. |
 | custom_rgxp        | Regular expression (`string`) | {{< version "4.11" >}} Custom regular expression to be used when using `'rgxp' => 'custom'` |
+| addWizardClass     | true/false (`bool`) | Prevents the `wizard` class from being added to the CSS. |
```
fiedsch: I would not say „prevent“ as this presumably corresponds to the `false` setting only. Maybe „Determines whether the wizard class …

fiedsch (changes requested):

docs/dev/reference/dca/fields.md
Tastaturberuf: Yes but by default it is set. I suggest "Remove the wizard class..." instead.

Tastaturberuf: Remove the `wizard` class in CSS if you use the wizard or picker option.

fiedsch: Sure. But my point was: this setting only removes the `wizard` class if set to `false`.
So if I can set an option to `true` or `false` the description can't be "it does something" (because it is "do something if set to `true`").

In addition: the described option name `add...` and the description "Prevents ... from being added" sound contradictory to me.

So maybe: 

```
| addWizardClass     | true/false (`bool`) | If `true`, add the `wizard` class (default). If `false`, do *not* add it. |
```

Tastaturberuf: "If `true`, add the `wizard` class (default)" is ultimately also not correct, but also only if a wizard is present. 

Then: "If `false`, do *not* add the `wizard` class, if a wizard is available.

fritzmg:
What about:

```
Whether or not to add the `wizard` CSS class for the widget if a wizard icon is in use. Default: `true`.
```
?

fritzmg:
I have updated the text now :). See also https://github.com/contao/contao/pull/3051

Thank you @Tastaturberuf

PR #720, opened 2021-05-03
Add Shortcode Tabbed-View Info

fkaminski:
...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,6 +16,7 @@ Only major versions are documented (e.g. Contao 4 and later on Contao 5).
 * Add line breaks after 80 characters in paragraphs.
 * Code examples should follow the Symfony Best Practices Book, use PHP 7.1 and
   put Contao related files to `contao`.
+* Use the »[tab view shortcode](https://learn.netlify.app/en/shortcodes/tabs/)« to display code examples.  
 * When using examples for PHP code or YAML configurations etc., include the example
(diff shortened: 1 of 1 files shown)


fritzmg:
Why do you want to add this? If you just want to show a single code example it makes no sense to put it in a tab?

Generally it would help if you provide more information for a Pull Request, not just "…" ;)

fkaminski:
> Why do you want to add this? If you just want to show a single code example it makes no sense to put it in a tab?

I thought now that this function is available it would be nice if everyone is informed about it

fkaminski:
> Generally it would help if you provide more information for a Pull Request, not just "…" ;)

OK (But this time there was nothing more to say than already in the headline) ...

fritzmg:
> I thought now that this function is available it would be nice if everyone is informed about it

You are trying to create a new rule though, rather than just informing people about a new feature of the theme. As such it does not make sense, since you should definitely not use a tab for a single code example (in my opinion).

The CONTRIBUTING.md does not list all the features of the theme, nor does it have to. It already says the following:

> See learn.netlify.com/en/cont/markdown/ and learn.netlify.com/en/shortcodes/ for available markdown and shortcode syntax.

fritzmg:
> But this time there was nothing more to say than already in the headline

I disagree, you always need to explain _why_ you want to make any change.

fkaminski:
Ok, understood.
Closed this.

PR #721, opened 2021-05-07
Fix urlSuffix notation in page-controllers.md

z3nFo:
Using urlSuffix="html" results in the path /foo/barhtml, which might not be desirable.


fritzmg:
Thank you @z3nFo :)

PR #723, opened 2021-05-11
Extended iflng docs

Toflar:
Before I translate this to English:

1. Does that make sense?
2. Should we repeat that tip for `ifnlng`?


netzarbeiter (approved):
Thank you, Yanick

netzarbeiter:
1. off course
2. just add "see iflng"

Toflar:
Ready then :)

PR #725, opened 2021-05-19
Add figure insert tag

fritzmg:
This adds the `{{figure::*}}` Insert Tag which was added in Contao 4.11, but still missing in the documentation. @m-vo anything to add?


m-vo:
The main difference of the figure insert tag is IMHO that you can leverage the full potential of the `FigureBuilder`. So you can not only define a size but also enable a lightbox, ship options and so on. 

Under the hood it uses the [FigureRenderer](https://github.com/contao/contao/blob/4.x/core-bundle/src/Image/Studio/FigureRenderer.php) that configures a figure builder instance. The query string of the insert tag  basically becomes the configuration parameter.

It also supports both PHP and Twig templates as the last parameter :smiley:. Maybe we should document this as well.

netzarbeiter:
Thank you, Fritz.
Moritz feel free to enhace the document.

PR #727, opened 2021-05-20
Change Indizierung to Indexierung

zonky2:
see first commet https://dict.leo.org/forum/viewGeneraldiscussion.php?idForum=4&idThread=1385507&lp=ende&lang=de

in my opinion Indexierung is more common...


netzarbeiter (approved):
Thank you, Ingolf

PR #728, opened 2021-05-21
Fix anchor link

stefansl:
The link to the section was broken.


fritzmg (approved):
There are probably more issues in this article, since it is only machine translated and thus not reviewed for such things. But there will be a hackathon in 2 weeks where we hopefully take care of most of these ;)

netzarbeiter (approved):
Thank you, Stefan

PR #729, opened 2021-05-22
Improve figure insert tag description

m-vo:
followup to #725 

I added some details to the `figure` insert tag description. Because there are quite a lot of options, I decided to alter the looks a bit. Let me know what you think about this. 

For now, this only contains the german version. I'll add the english translation once we agree this is the way to go. 

image


fritzmg (commented):
Yeah, kinda deserves its own section may be. It's a bit cumbersome to manage this within a markdown table cell. But gets my approval as is :). Can you may be do the English version as well?

m-vo:
Yeah, I agree. Maybe that's something for the docs hackaton. :slightly_smiling_face: 

> Can you may be do the English version as well?

Done in 287879f90fed7f6762eddaeac6b5d8768f7233ae.

netzarbeiter:
Thank you, Moritz

PR #731, opened 2021-05-25
Update edit-data-records.en.md

stefansl:
The machine translation was "overwrite", in Contao it's always called "override" ... so I fixed this + new screenshots (english + retina size)


netzarbeiter (approved):
Thank you, Stefan

Issue #687, opened 2021-03-16
Improve Core Widget and DCA field documentation


bezin:
Hi there, 

I really appreciate the (not so new anymore) documentation and it was a huge step forward 🎉 

One thing that is still a little hard to understand is the [DCA field documentation](https://docs.contao.org/dev/reference/dca/fields/). 

What we currently have is a list of all available options, structured in a way that does not help you if you are new to Contao, and – honestly – only even helps you a little if you are already a few years in. The key issue is IMO, that nowhere in the documentation we actually show **which widget can be configured with which options**. We know all the options, we know all available Core widgets, but there is no matching between them. 

As a positive example I would like you to have a look at the [field documentation of the Kirby CMS](https://getkirby.com/docs/reference/panel/fields/checkboxes#field-properties). There is a whole page dedicated to every single field (Kirbys widget) with all the properties that can be set to configure this particular field. You also find plenty examples of different configurations, but I think we all would love to see more examples everywhere in the documentation ;-)

There are some issues with this approach for sure: In Slack, @ausi already mentioned,that one would need to document general options that apply to many or all widgets in a way that does not turn into an unmaintainable mess. 

Since I think it is easier to reason about this if there is something to reason about, I'll document a widget as I would like to see them and we can further discuss it from there :-) Comments and thoughts much appreciated. 

Cheerio

bezin:
Hello again,

I „quickly“ documented the `checkbox` widget as a very first, totally unfinished draft to show what I would like as a result in the end. 

As you see, my main focus would be on how to use this specific widget. The _general options_ table with options shared among all widgets is an issue editing-wise: Maybe there is a HUGO way to share content across multiple pages? It would be nice to edit it in a single place, but reference it on every page. The _widget specific options_ are also shared between some widgets, but I feel like these should be explained in the context of a widget.

![Screenshot_2021-03-16 Checkbox Contao Developer Documentation](https://user-images.githubusercontent.com/4400435/111338693-6ba2e580-8677-11eb-98c8-77012a4391aa.png)

Source is here: https://github.com/bezin/contao-docs/tree/widget-reference

BTW: The Kirby documentation is self-documenting when it comes to the field config, as the Kirby panel (back-end) is implemetend in Vue and the documentation just reads the component definition 🤷

fritzmg:
Looks very good 👍 

Now if you could provide a PR containing _all_ the widgets that would be super-duper 😁 

[...]

PR #732, opened 2021-05-27
Add widget reference and document checkbox widget

bezin:
This adds a widget reference as sketched out in #687. As discussed, this includes a blank page for every widget, so any contributor is welcome :-). 

The checkbox widget is already documented – though still not perfect, it can work as a template for other widgets.

Feel free to comment :-)


fritzmg:
Thank you @bezin for this awesome contribution :). I put the notice for each WIP widget article in a short code, so that we don't have to duplicate it.

fritzmg (approved):

PR #735, opened 2021-06-03
Update migrations.md

Tastaturberuf:
Use nice shorthand method

I think the documentation should show whats part of the framework and its methods.

https://github.com/contao/contao/blob/13ca9de98fdef79992a094d962ae9297138a8b24/core-bundle/src/Migration/AbstractMigration.php#L22


fritzmg (approved):

fritzmg:
Thank you @Tastaturberuf :)

PR #740, opened 2021-06-03
Improve manage template section (hackaton, task 28)

m-vo:
(also includes some changes to the german version)


netzarbeiter (approved):
Thank you, Moritz

PR #741, opened 2021-06-03
Update mashine translated text

fiedsch:
Changed in german doc also:
* meta files are obsolete. Should be metadata
* remove the sentence "In the following, the individual input fields will be explained to you." as we don't have this in the other sections


netzarbeiter (approved):
Thank you, Andreas

PR #747, opened 2021-06-03
Improve template data section (hackaton, task 30)

m-vo:
(also includes some changes to the german version)

Removed the reference to `dump($this)` as it's IMHO less helpful to give both options and added a word about the output landing in the debug toolbar instead.

### DIFF
--- a/docs/manual/layout/templates/template-data.de.md
+++ b/docs/manual/layout/templates/template-data.de.md
@@ -8,25 +8,19 @@ weight: 50
 ---
 
 Die verfügbaren Template-Daten variieren je nach Quelle der Vorlage. In der Regel ist der vollständige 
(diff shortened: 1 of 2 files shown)


fritzmg:
Well, `dump($this)` has an advantage: it will show you from which template the debug output comes from. `$this->dumpTemplateVars` does not. But may be that is something that the core can improve on.

m-vo:
True, but it also contains lots of other unnecessary things (besides `$arrData`). IMHO we should rather improve the `dumpTemplateVars` output or simply always add the template name to the context.

fritzmg (approved):

netzarbeiter (approved):
Thank you, Moritz

PR #752, opened 2021-06-03
Improve template inheritance section (hackaton, task 31)

m-vo:
(as always: this also tries to improve the German version)


netzarbeiter (approved):
Thank you, Moritz

PR #754, opened 2021-06-03
Install Contao (Hackathon task 4)

stefansl:
Some changes in the german docs were necessary too. I committed them seperatly ...


netzarbeiter (approved):
Thank you, Stefan

PR #755, opened 2021-06-03
Isotope core

netzarbeiter:


docs/manual/extensions/contao-easy_themes.en.md
```diff
@@ -5,55 +5,56 @@ aliases:
...
-{{% /notice %}}
-
 **[terminal42/contao-easy\_themes](https://packagist.org/packages/terminal42/contao-easy_themes)**
 
 *from [terminal42 gmbh](https://www.terminal42.ch/de/)*
 
-Enables direct access to stylesheets, modules, page layouts and image  
- sizes - not only in case of multiple themes a click number reducer.
+Enables direct access to stylesheets, modules, page layouts and image sizes - not only in case of multiple themes a 
+click number reducer.
```
Toflar, line 13: I don't think "click number reducer" is an expression.
What about "reduces the number of clicks even if only one theme is used"?

Toflar (changes requested):
This PR has to be updated to the upstream main branch (it includes lots of unrelated changes) :)

PR #758, opened 2021-06-03
About Contao Manager (Task No. 5)

stefansl:
As the hosting provider setting is not needed anymore, I removed it in DE and EN. I also updated the german screenshots


netzarbeiter (approved):
Thanky you, Stefan

PR #768, opened 2021-06-04
Improve move contao section (hackaton, task 8)

m-vo:
This improves the "Move Contao" section.

* I wasn't really happy with the original document so I restructured it, made use of the tab feature and removed redundant content (see installation).
* I also adjusted the args of the dump/import commands because they were IMO easy to get wrong (no space in `-uUsername` for instance) or potentially generating bad files (missing `--hex-blob`).

These two things should get reviewed first IMO, because we likely want to port them to the German version as well.

Images will be adjusted by @stefansl in a followup PR.

### DIFF
--- a/docs/manual/installation/move-contao.en.md
+++ b/docs/manual/installation/move-contao.en.md
@@ -23,7 +23,7 @@ To reduce the risk of conflicts, make sure your source and target server both ru
 You can either create a MySQL dump with the graphical database administration tool [phpMyAdmin](https://www.phpmyadmin.net/)
 or use the `mysqldump` program from the command line.
 
-{{< tabs >}}
+{{< tabs groupId="mysql-transfer" >}}
 {{% tab name="phpMyAdmin" %}}
 Log into "phpMyAdmin", select the database you want to export, select the "Export" tab in the upper menu and click "Ok".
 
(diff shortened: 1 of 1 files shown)


docs/manual/installation/move-contao.en.md
```diff
@@ -1,125 +1,102 @@
...
 {{% /notice %}}
 
-## Move Contao with the Contao Manager
+## Transferring the database
+### Export the database (source)
+You can either create a MySQL dump with the graphical database administration tool [phpMyAdmin](https://www.phpmyadmin.net/)
+or use the `mysqldump` program from the command line.
 
-### Export database on the local server
+{{< tabs >}}
```
m-vo: @fritzmg Would this need a tab group id? If so, how would this look?

(edit: I added one, as it seems necessary: https://learn.netlify.app/en/shortcodes/tabs/)

docs/manual/installation/move-contao.en.md
```diff
@@ -1,125 +1,102 @@
...
+application related files.
 
-Moving a local installation to a live server is almost the same as a new installation.
+1. [Transferring the database](#transferring-the-database)
+2. [Transferring the files](#transferring-the-files)
+3. [Installing Contao](#installing-contao)
 
 {{% notice warning %}}
-To avoid inconvenience during the move, your local server should have the **[same PHP version](/en/installation/system-requirements/#minimum-php-requirements)** as running on the live server.
+To reduce the risk of conflicts, make sure your source and target server both run the **[same PHP version](/en/installation/system-requirements/#minimum-php-requirements)**.
```
netzarbeiter:
```suggestion
To reduce the risk of conflicts, make sure your source and target server both run the **[same PHP version](../system-requirements/#minimum-php-requirements)**.
```

netzarbeiter (changes requested):
Thank you, Moritz. Can you please adjust the german version too.

docs/manual/installation/move-contao.en.md
m-vo: adjusted in 2570672

netzarbeiter (approved):

PR #771, opened 2021-06-07
Followup to 768

m-vo:
@netzarbeiter #768 got merged to quickly (german version + CS) - here are the missing commits rebased on current `main`.


netzarbeiter (approved):
Thank you, Moritz

PR #783, opened 2021-06-27
Add chapter about Security

fritzmg:
This adds a basic chapter about the usage of the Symfony Security Component within Contao. It explains some basic principles and links to the appropriate documentation within Symfony. It also mentions the available security voters of Contao and describes how to extend back end user permissions. It also documents the new `contao_group` permissions attribute of https://github.com/contao/contao/pull/3124

@aschempp can you glance over it to check for factual errors?

### DIFF
--- a/docs/dev/framework/security.md
+++ b/docs/dev/framework/security.md
@@ -84,16 +84,19 @@ If you want to learn more about Symfony's Security Component use the provided li
 implmentation details that are unique to Contao.
 
 Since within Contao you can put a login form on basically any page, Contao does not utilise Symfony's built-in 
-[`form_login` Authentication Provider][SymfonyFormLogin]. Instead, Contao implements its own user providers, 
-[user checkers][SymfonyUserChecker] and [request matcher service][SymfonyRequestMatcherService], the latter of which checks the 
-[request scope][RequestScope]. For example in the front end, all URLs to Contao pages will have the `_scope` request attribute set to 
-`frontend` and the `contao_frontend` firewall will thus be applicable to all these URLs.
+[`form_login` Authentication Provider][SymfonyFormLogin]. Instead, Contao implements its own [user checker][SymfonyUserChecker] and 
+[request matcher service][SymfonyRequestMatcherService], the latter of which checks the [request scope][RequestScope]. For example in the 
+front end, all URLs to Contao pages will have the `_scope` request attribute set to `frontend` and the `contao_frontend` firewall will thus 
+be applicable to all these URLs. Contao implements an [authentication listener][SymfonyAuthenticationListener] which will check for any
+POST request containing the parameters `username` and `password` and the parameter `FORM_SUBMIT` with the value `tl_login` (as these are the
+parameters used by Contao's login module).
 
 
 ## Voters
 
-Starting with Contao **4.7** Contao implements so called [Voters][SymfonyVoters] in order to easily check whether an authenticated user is
-authorized to access specific resources. These voters are automatically used by Symfony via the [Security Helper][SecurityHelperService].
+Starting with Contao **4.7** Contao implements [Voters][SymfonyVoters] in order to easily check whether an authenticated user is authorized 
+to access specific resources. These voters are automatically added to Symfony's security system and then invoked when the respective
+permission is accessed via the [Security Helper][SecurityHelperService].
 
 {{< version "4.7" >}}
 
@@ -126,11 +129,11 @@ $security->isGranted('contao_user.can_edit_page', $pageModel);
 
 {{< version "4.12" >}}
 
-The security helper can also be used to check whether a front end or back end user belongs to any of the specified user groups:
+The security helper can also be used to check whether a front end user belongs to any of the specified user groups:
 
 ```php
-$security->isGranted('contao_group', $groupId);
-$security->isGranted('contao_group', [/* array of group IDs */]);
+$security->isGranted('contao_member.groups', $groupId);
+$security->isGranted('contao_member.groups', [/* array of group IDs */]);
 ```
 
 {{% notice tip %}}
@@ -164,23 +167,34 @@ $security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES);
 
 To implement your own back end access rights (e.g. for custom modules in the back end) the following steps are necessary:
 
-1. Add your new permission to `$GLOBALS['TL_PERMISSIONS']`.
+1. Add your new permission to `$GLOBALS['TL_PERMISSIONS']`. This registers this permission to be used by Contao's `contao_user` voter.
 2. Add your new permission to `tl_user` and `tl_user_group`.
 
 ```php
 // contao/config/config.php
-$GLOBALS['TL_PERMISSIONS'][] = 'my_permission';
+$GLOBALS['TL_PERMISSIONS'][] = 'my_permissions';
 ```
 
 ```php
 // contao/dca/tl_user.php
 use Contao\CoreBundle\DataContainer\PaletteManipulator;
 
-$GLOBALS['TL_DCA']['tl_user_group']['fields']['my_permission'] = [
+$GLOBALS['TL_DCA']['tl_user_group']['fields']['my_permissions'] = [
     'exclude' => true,
     'inputType' => 'checkbox',
-    'sql' => ['type' => 'boolean', 'default' => false],
+    'eval' => ['multiple' => true],
+    'options' => [
+        'first_permission' => 'First permission',
+        'second_permission' => 'Second permission',
+    ],
+    'sql' => ['type' => 'blob', 'notnull' => false],
 ];
+
+	'exclude'                 => true,
+	'inputType'               => 'checkbox',
+	'foreignKey'              => 'tl_calendar.title',
+	'eval'                    => array('multiple'=>true),
+	'sql'                     => "blob NULL"
 
 PaletteManipulator::create()
     ->addLegend('my_legend', null)
@@ -191,12 +205,17 @@ PaletteManipulator::create()
 ```
 
 ```php
+// contao/dca/tl_user_group.php
 use Contao\CoreBundle\DataContainer\PaletteManipulator;
 
-$GLOBALS['TL_DCA']['tl_user_group']['fields']['my_permission'] = [
+$GLOBALS['TL_DCA']['tl_user_group']['fields']['my_permissions'] = [
     'exclude' => true,
     'inputType' => 'checkbox',
-    'sql' => ['type' => 'boolean', 'default' => false],
+    'options' => [
+        'first_permission' => 'First permission',
+        'second_permission' => 'Second permission',
+    ],
+    'sql' => ['type' => 'blob', 'notnull' => false],
 ];
 
 PaletteManipulator::create()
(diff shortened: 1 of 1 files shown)


docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+description: Contao's Symfony Security implementation.
+aliases:
+    - /framework/security/
+---
+
+Contao uses [Symfony's Security Component][SymfonySecurityComponent] to handle front end and back end user authentication and authorization.
+The Contao Managed Edition provides its own [firewall][SymfonyFirewall], [user providers][SymfonyUserProvider] and 
+[access control][SymfonyAccessControl] via the `contao/manager-bundle`. If you do not use the Managed Edition and added Contao to your
+custom Symfony application, you will have to register Contao's security settings yourself, as mentioned in the 
+[Getting Started article][ContaoConfiguration]. Contao's Symfony Security configuration [in Contao **4.9**][Contao49SecurityYaml] looks like this for example:
```
aschempp, line 12: not sure it's a good idea to duplicate this here again. Maybe it should be referenced to the actual file instead?

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+        - { path: ^/contao/logout$, roles: IS_AUTHENTICATED_ANONYMOUSLY }
+        - { path: ^/contao(/|$), roles: ROLE_USER }
+        - { path: ^/, roles: [IS_AUTHENTICATED_ANONYMOUSLY] }
+```
+{{% /expand %}}
+
+If you want to learn more about Symfony's Security Component use the provided links to read up on. This documentation will only cover
+implmentation details that are unique to Contao.
+
+Since within Contao you can put a login form on basically any page, Contao does not utilise Symfony's built-in
```
aschempp, line 86: `form_login`, `request matcher`, `user provider` and `user checker` are not related to each other at all.

1. Request matchers allow us to not match on paths but on the scope attribute (anything because request is matched through a service). The idea behind this was also to allow any other custom paths in your Symfony app that won't be affected by the Contao firewall.

2. form_login is the default Symfony username/password factory, but it does not work for Contao because we don't have known login paths (specifically in the front end). We match on `FORM_SUBMIT = tl_login` instead, and since we have our own firewall factory we do a bunch of other stuff as well (mostly hardcoding stuff instead of having a firewall configuration).

3. user provider is something you always need with a firewall, so not really Contao-specific

4. user checker is Symfony-default as well. it _could_ be optional but having one is pretty standard, you can e.g. check if a user has been disabled to throw him out of an active login session.

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+[`form_login` Authentication Provider][SymfonyFormLogin]. Instead, Contao implements its own user providers, 
+[user checkers][SymfonyUserChecker] and [request matcher service][SymfonyRequestMatcherService], the latter of which checks the 
+[request scope][RequestScope]. For example in the front end, all URLs to Contao pages will have the `_scope` request attribute set to 
+`frontend` and the `contao_frontend` firewall will thus be applicable to all these URLs.
+
+
+## Voters
+
+Starting with Contao **4.7** Contao implements so called [Voters][SymfonyVoters] in order to easily check whether an authenticated user is
+authorized to access specific resources. These voters are automatically used by Symfony via the [Security Helper][SecurityHelperService].
```
aschempp: Not sure what your definition of "automatically used" is. If noone calls `isGranted` then they are obviously not used.

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+$security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES);
+```
+{{% /notice %}}
+
+
+## Custom Back End Access Rights
+
+To implement your own back end access rights (e.g. for custom modules in the back end) the following steps are necessary:
+
+1. Add your new permission to `$GLOBALS['TL_PERMISSIONS']`.
```
aschempp: maybe we need to explain what this does?

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+PaletteManipulator::create()
+    ->addLegend('my_legend', null)
+    ->addField('my_permission', 'my_legend', PaletteManipulator::POSITION_APPEND)
+    ->applyToPalette('extend', 'tl_user')
+    ->applyToPalette('custom', 'tl_user')
+;
+```
+
+```php
+use Contao\CoreBundle\DataContainer\PaletteManipulator;
```
aschempp, line 209: probably missing file name here

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+];
+
+PaletteManipulator::create()
+    ->addLegend('my_legend', null)
+    ->addField('my_permission', 'my_legend', PaletteManipulator::POSITION_APPEND)
+    ->applyToPalette('default', 'tl_user_group')
+;
+```
+
+Once that is done you can check for this permission in your own back end controller for example. These permissions can be checked via the
```
aschempp, line 228: unfortunately, if I remember correctly, you cannot have a single boolean field as a permission (inheritance (`TL_PERMISSION`) doesnt work with that). It has to be a multiple-checkbox that can be `array_merge`d

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+            throw new AccessDeniedException('Not enough permissions to access this controller.');
+        }
+
+        return new Response($this->twig->render('my_backend_route.html.twig', []));
+    }
+}
+```
+
+{{% notice note %}}
+We do not need to check whether a user is logged in this case, since the Contao firewall automatically checks this
```
aschempp, line 271: but `ROLE_ADMIN` does that check as well

docs/dev/framework/security.md
fritzmg: It already links to the actual file ;). We duplicate other things as well (like the config reference) so I think it's fine to have this here, to give a quick overview.

docs/dev/framework/security.md
fritzmg: Ad 1) & 2): that's what I want to say with that sentence. Regarding 3) and 4) I'll flesh this out more.

docs/dev/framework/security.md
fritzmg: Meaning that nobody calls your voter directly, it will be automatically called when using `isGranted` of the security helper, if the given attribute is handled by your voter.

docs/dev/framework/security.md
fritzmg: It does work as far as I tested (see the example below).

docs/dev/framework/security.md
fritzmg: Yes, but this is a check for _not_ admin, which would be true for an unauthenticated user - but due to the firewall we don't have to check for that.

docs/dev/framework/security.md
aschempp: ok… I would find that rather confusing. _Obviously_ a voter is supposed to be used if you register one, the term _automatically_ implies (to me) that I don't have to check permissions if I implement a voter. Maybe something like _You can use the Contao voters in your code to check permissions._

fritzmg: But you never use a voter yourself? I would find the latter confusing. You implement a voter and then Symfony tags the voter's service and injects it elsewhere, so that they are invoked whenever `isGranted` is called.

aschempp: that's true, it would rather be _use Symfony security to check permissions_ then?

docs/dev/framework/security.md
aschempp: I don't think it does 😉

aschempp: Not sure it does. The result is always an array for sure, which looks pretty confusing:
https://github.com/contao/contao/blob/4.x/core-bundle/src/Resources/contao/classes/BackendUser.php#L446-L447

fritzmg: Yes, and this results in the following check in `hasAccess`: https://3v4l.org/jhoYa

And thus it will return true.

docs/dev/framework/security.md
fritzmg: I cannot find a way to write a sentence without _automatically_ 😁. For me, Symfony automatically uses these Voters. All I do is create the class and nothing else (with autowiring and autoconfiguration enabled). I don't have to tell Symfony anything to use the voter, Symfony uses them _automatically_ when checking for permissions.

docs/dev/framework/security.md
fritzmg: Yeah, I would add that adding your permission field to `TL_PERMISSIONS` enables Contao's `contao_user` voter to check for this permission.

docs/dev/framework/security.md
aschempp: Symfony _will_ add the voter to the security system "automatically". But it will not check for any permission related to the new voter, if noone will ask for it. And asking for a permission your voter added is always your job.

docs/dev/framework/security.md
aschempp: it _might_ work, yes. I remember I had issues with this, probably in Contao 2 or 3. It will however add confusing data, since the user object will now hold an array of this field, even though the user field is a checkbox only.
I'm just saying, maybe it's better to make an example of what we use everywhere (multiple checkboxes) instead of just one 😊

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+$security->isGranted('contao_user.can_edit_fields', 'tl_page');
+
+// whether user can edit the given page (array of row or page model)
+$security->isGranted('contao_user.can_edit_page', [/* row of page data */]);
+$security->isGranted('contao_user.can_edit_page', $pageModel);
+```
+
+{{< version "4.12" >}}
+
+The security helper can also be used to check whether a front end or back end user belongs to any of the specified user groups:
```
mroi: It seems not true that the security helper can be user for backend users. I understand it’s limited to frontend users.

docs/dev/framework/security.md
```diff
@@ -0,0 +1,275 @@
...
+
+```php
+$security->isGranted('contao_group', $groupId);
```
mroi: The string constant in `ContaoCorePermissions` is `'contao_member.groups'`, not `'contao_group'`.

docs/dev/framework/security.md
fritzmg: It can be used for both.

mroi: I tried to query group membership of backend users and this stopped me:
https://github.com/contao/contao/blob/4.x/core-bundle/src/Security/Voter/MemberGroupVoter.php#L39-L41
… but it’s possible I’m doing something wrong.

fritzmg: `contao_member.groups` is only for front end users. The security helper in general can be used both for front and back end users otherwise. But yes, the implementation of this voter has changed since this PR was first drafted so the wording needs to be changed accordingly.

@aschempp I don't remember the reasoning why we cannot vote on back end user groups as well now - what was the reason again?  😁

aschempp: Because the back end groups are not relevant. They inherit permissions to the user object, which can then be checked upon.

fritzmg:
Reminder to myself: redirect https://docs.contao.org/books/cookbook/de/Benutzer-Authentifizierung-und-Rechtepruefung.html#aufruf-des-benutzer-objektes-im-frontend to the new article.

fritzmg:
@aschempp rewrote some parts and merged it for now. Feel free to suggest any changes :)

PR #784, opened 2021-06-28
Documented the new ResponseContext introduced in 4.12

Toflar:
I did not document every single method in `ResponseContext` or the `JsonLdManager`.
I think the real value of the docs is to explain the concepts. Answering the question of "why is this?".
Developers will figure out what the other methods on the `ResponseContext` do and in case there are uncertainties, we can always get more detailed in the docs :)

### DIFF
--- a/docs/dev/framework/response-context.md
+++ b/docs/dev/framework/response-context.md
@@ -24,7 +24,7 @@ In Contao, however, most of the content is dynamically created. That's why it's 
 System after all. Let's take e.g. a news reader front end module as an example. It displays the 
 detail of a certain news entry and thus also would like to adjust the `<title>` tag in the `<head>`
 section of the HTML document. The controller, however, does not know if that news reader front end module
-was placed on that page and the front end module, on the other hand, does not even know if there is a `<title>`
+was placed on that page and the front end module on the other hand does not even know if there is a `<title>`
 element it can update.
 The news reader element is a fragment only so it can be used in a completely different context other than a
 regular HTML web page. E.g. it could be rendered as an ESI fragment, a partial loaded via Ajax, rendered into
--- a/docs/dev/framework/response-context.md
+++ b/docs/dev/framework/response-context.md
@@ -27,7 +27,7 @@ section of the HTML document. The controller, however, does not know if that new
 was placed on that page and the front end module on the other hand does not even know if there is a `<title>`
 element it can update.
 The news reader element is a fragment only so it can be used in a completely different context other than a
-regular HTML web page. E.g. it could be rendered as an ESI fragment, a partial loaded via Ajax, rendered into
+regular HTML web page. E.g. it could be rendered as an ESI fragment, a partial loaded via Ajax or rendered into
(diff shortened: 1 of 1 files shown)


docs/dev/framework/response-context.md
```diff
@@ -0,0 +1,174 @@
...
+This section describes the "Response Context" concept. A concept that is special
+to Contao and thus deserves its own section in the documentation.
+
+## Motivation
+
+Content Management Systems do not share all the same properties as static site generators, or your
+good old Symfony based website.
+In a traditional Symfony application, you as the developer control 100% of the output of
+a particular HTML website. Speaking in Symfony terms, your controller 
+controls the full `Response` instance, from its HTTP headers to it's content, everything from
```
fritzmg:
```suggestion
controls the full `Response` instance, from its HTTP headers to its content, everything from
```

docs/dev/framework/response-context.md
```diff
@@ -0,0 +1,174 @@
...
+the opening `<html>` to the closing `</html>`, all the scripts, the stylesheets, the page title and
+the description.
+
+In Contao, however, most of the content is dynamically created. That's why it's a Content Management
+System after all. Let's take e.g. a news reader front end module as an example. It displays the 
+detail of a certain news entry and thus also would like to adjust the `<title>` tag in the `<head>`
+section of the HTML document. The controller, however, does not know if that news reader front end module
```
fritzmg, line 26:
```suggestion
section of the HTML document. The controller however does not know if that news reader front end module
```

docs/dev/framework/response-context.md
```diff
@@ -0,0 +1,174 @@
...
+was placed on that page and the front end module, on the other hand, does not even know if there is a `<title>`
```
fritzmg:
```suggestion
was placed on that page and the front end module on the other hand does not even know if there is a `<title>`
```

docs/dev/framework/response-context.md
```diff
@@ -0,0 +1,174 @@
...
+element it can update.
+The news reader element is a fragment only so it can be used in a completely different context other than a
+regular HTML web page. E.g. it could be rendered as an ESI fragment, a partial loaded via Ajax, rendered into
```
fritzmg:
```suggestion
regular HTML web page. E.g. it could be rendered as an ESI fragment, a partial loaded via Ajax or rendered into
```

fritzmg (changes requested):
Thank you very much :). Content looks good, just some minor grammar points.

docs/dev/framework/response-context.md
Toflar: Are you sure? I was taught to separate `however` with commas.

fritzmg: Yeah, you are right. Even when it is used for emphasis, commans are usually put around it, though apparently there is no clear rule. So leave it in ;)

fritzmg (approved):

fritzmg:
Thank you @Toflar :)

PR #786, opened 2021-07-01
Add requirements + example code for SQL-table

ACCakut:
Adding a clarification and sample code to create a suitable SQL-table for saving forms in it.


fiedsch (changes requested):
beforehand (e.g. via phpMyAdmin or as [DCA](../../../../dev/reference/dca/)) in order to show up as a choice. The SQL-table

fiedsch (changes requested):
The SQL-table -> The database table

netzarbeiter (approved):
Thank you, ACCakut

PR #787, opened 2021-07-01
Add requirements + example code for SQL-table

ACCakut:
Adding a clarification and sample code to create a suitable SQL-table for saving forms in it.


netzarbeiter (approved):
Thank you, ACCakut

PR #788, opened 2021-07-01
Update filter.en.md

berecont:
in line 127 'filtr-item DATA-dog' missing hyphen between DATA and dog. So the query in file block_searchable_filter.html5 will work.


netzarbeiter:
Thank you, Bernhard

PR #794, opened 2021-07-20
Fix missing nullable type in onload callback

koertho:
This PR fix a missing nullable type in the onload callback example.


fritzmg:
That's not really necessary: https://3v4l.org/plHfN

koertho:
Ok, wasn't aware that PHP accept this type of declaration. I would still prefer the ?, as it makes the syntax more clear, but if you disagree, you can close this PR.

fritzmg:
I think it's superfluous. There are also numerous places in the Contao Core as well as Symfony where an optional parameter is defined with `= null` without `?` on the type declaration.

PR #797, opened 2021-07-30
Update file-manager.en.md

bezin:
Fix issues with icons and add paragraph about updating existing files from the file manager.


netzarbeiter (approved):
Thank you, Benedict

PR #798, opened 2021-07-30
Update file-manager.de.md

bezin:
Add what happens if you upload a file with a filename that already exists


netzarbeiter (approved):
Thank you, Benedict

PR #809, opened 2021-08-18
Fixed new example

Toflar:
Can't we use prepared statements here? At least for `type` we can and should for sure.


fritzmg:
Imho it's not necessary for `type` since it's a constant. And for the `$ids` it's also not necessary since the IDs will already be forced to integers. The core uses it the same way. Also I think your example would not work, because Contao would then create the following query: `id IN ('1,2,3')`.

Toflar:
Yeah right, we should probably still have `type = ?` to remain consistent with the rest?

fritzmg:
Yes, I am fine with passing `type` as a parameter :)

Toflar:
Voilà :)

fritzmg (approved):

PR #821, opened 2021-09-03
Fix file hint

z3nFo:
The file hint within the code example was pointing to contao/dca/tl_module.php. In the example a tl_content palette is added, which should be placed in contao/dca/tl_content.php


fritzmg (approved):

fritzmg:
Thank you @z3nFo :)

PR #836, opened 2021-09-17
Update back-end-routes.md

doishub:
Change `/templates` to `/views`.


fritzmg:
That only applies to bundles, doesn't it? For the App, `templates/` should be the correct folder.

doishub:
Oh, that may be. I have just tested it in a bundle. Maybe a note should be added here, or is this guide only for development within a project?

fritzmg:
Actually the new standard is `templates/` even for bundles. See https://symfony.com/doc/current/bundles/best_practices.html#directory-structure (same for 4.4: https://symfony.com/doc/4.4/bundles/best_practices.html)

PR #781, opened 2021-06-25
Guide - Handling/Saving Form Data with Core Tools or Leads Extension

fkaminski:
See #775
Initial DE / EN translation after review.

PR #841, opened 2021-09-24
Simplify + optimize code example / 'handling/saving form data' section

m-vo:
followup to #781

I applied some best practices to the code example:
* `COUNT` should be preferred over fetching records (cheaper)
* `fetchOne` should be preferred over `fetchAssociative` if values are not needed (cheaper)
* methods should be `private` by default
* the `duplicateCheck` method is redundant - either the `slugExists` should be referenced as callback method or (what I did) the method can be inlined

### DIFF
--- a/docs/manual/guides/form-save.de.md
+++ b/docs/manual/guides/form-save.de.md
@@ -89,8 +89,8 @@ use Doctrine\DBAL\Connection;
 class PrepareFormDataListener
 {
     // Change these variables for your form, calendar and author
-    private const FORM_ID = 6;
-    private const CALENDAR_ID = 2;
+    private const FORM_ID = 3;
+    private const CALENDAR_ID = 5;
     private const AUTHOR_ID = 1;
 
     private $slug;
--- a/docs/manual/guides/form-save.de.md
+++ b/docs/manual/guides/form-save.de.md
@@ -89,8 +89,8 @@ use Doctrine\DBAL\Connection;
 class PrepareFormDataListener
 {
     // Change these variables for your form, calendar and author
-    private const FORM_ID = 3;
-    private const CALENDAR_ID = 5;
+    private const FORM_ID = 6;
+    private const CALENDAR_ID = 2;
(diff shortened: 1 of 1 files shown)


fritzmg:
Did you check if the code works at all by any chance? There are reports on the forum that it may be does not 😁

m-vo:
> Did you check if the code works at all by any chance? There are reports on the forum that it may be does not grin

Lol, no. Refactored this in the GitHub IDE and just run it in the brain-executable. What is not working what was working before, though?

fritzmg:
https://community.contao.org/de/showthread.php?81705-Formulardaten-f%C3%BCr-Event-speichern

I already refactored the code once, but may be I broke something while doing so.

m-vo:
Just guessing: Is it that `strtotime` returns `false` in the case of an error and Contao converts this to `''` somewhere?

fritzmg:
I'll do some testing :) (with your updated code)

fritzmg:
Ok, so - in case `strtotime` returns `false`, this is actually not a problem, since `false` seems to be handled equivalent to `0` in that case. So I am not sure what happens in the case of the forum user.

We should however adjust the code to

```php
strtotime(trim($submittedData['startDate'])) ?: null
```

Apart from that I discovered an issue in Contao.  `\Contao\Form` uses

```php
Widget::getEmptyValueByFieldType($GLOBALS['TL_DCA'][$this->targetTable]['fields'][$k]['sql']);
```

to get the empty field value, in case the value is an empty string (and only then, so this will not be executed if it is `false`). However, the DCA for the target table is never loaded, so even though `tl_calendar_events.endDate` for example only accepts integers and `null`, an empty string will not be converted to `null`.

fritzmg:
PR for the bug: https://github.com/contao/contao/pull/3499

fritzmg (approved):

netzarbeiter (approved):
Thank you, Moritz

m-vo:
Thanks for debugging @fritzmg ❤️

PR #856, opened 2021-10-09
Explain TRADITIONAL SQL mode

fritzmg:
This explains the `TRADITIONAL` SQL mode setting, which the Contao Install Tool currently suggests should be used (in case the strict mode is not enabled by default), in more detail. See also https://github.com/contao/contao/pull/3554


netzarbeiter (approved):
Thank you, Fritz

PR #861, opened 2021-10-22
Add doNotRedirectEmpty and fix some other things

fritzmg:
This PR provides the following:

* Added the missing `doNotRedirectEmpty` `localconfig` setting.
* Fixed the notices for the front end configuration in the German version.
* Added the missing notices for the front end configuration in the English version.
* Moved the cron job settings in the English version (which was already done in the German version).


netzarbeiter (approved):
Thank you, Fritz

PR #874, opened 2021-11-11
Delete Contao call

zonky2:


netzarbeiter (approved):
Thank you, Ingolf

netzarbeiter:
https://github.com/contao/docs/commit/ec9f20667ca01fdfe163d4bc6dacc641d25a83b2

fiedsch:
For a reader that does not know the history of how Contao calls where held, the question might arise: "what does Mumble have to do with Contao and why do we have it in the Contao documentation?"

A note like "Mumble was used to hold the Contao calls until replaced by Slack huddles" right at the start of the text might clear this.  

What do you think?

netzarbeiter:
Thank you for your feedback.

Don't you think the note on the top of the page does the trick?

fiedsch:
Yes, sure. I was only reading the diff showing @zonky2's changes and did not see the note. My bad :-(

PR #876, opened 2021-11-17
Change default from boolean

zonky2:
boolean is TINYINT and strict mode send error without default value

### DIFF
--- a/docs/dev/reference/widgets/checkbox.md
+++ b/docs/dev/reference/widgets/checkbox.md
@@ -57,7 +57,7 @@ If you simply want to toggle a property:
     'inputType' => 'checkbox',
     'sql' => [
         'type' => 'boolean',
-        'default' => false,
+        'default' => 0,
     ],
 ],
 // ...


fritzmg:
Hm, I cannot reproduce. What exact error are you getting?

zonky2:
`General error: 1364 Field 'xy' doesn't have a default value`

fritzmg:
There must be a mistake on your end. When using

```php
// contao/dca/tl_content.php
$GLOBALS['TL_DCA']['tl_content']['fields']['foobar'] = [
    'sql' => [
        'type' => 'boolean',
        'default' => false,
    ],
];
```

then `contao:migrate` will execute

```
ALTER TABLE tl_content ADD foobar TINYINT(1) DEFAULT '0' NOT NULL
```

zonky2:
ok. - I´ll check this

fritzmg:
May be the field was already present in your database? But even then: the default value change should be picked up by Doctrine and thus change during `contao:migrate`.

PR #877, opened 2021-11-18
2 Hinweise für contao:crawl ergänzt

contaoacademy:


docs/dev/reference/commands.md
```diff
@@ -112,6 +112,10 @@ The options are far more important so let's get to them right away:
 | `--enable-debug-csv` | By default the subscriber results are written to the standard output. You can ask the command to write everything to a CSV file by passing this option. |
 | `--debug-csv-path`   | This option allows you to override the default CSV file path, if you used `--enable-debug-csv`. |
 
+**Notes**
+1. If you have problems with the crawler, check if you have configured a domain on your root page.
```
fritzmg: That's a little too vague in my opinion. For the crawler to be able to generate the correct URLs, the domain must be defined somewhere. So you either have to define the domain in the website root's DNS settings - or, if you want/need to keep it empty for whatever reason, you at least need to define the `%router.request_context.host%` and `%router.request_context.scheme%` parameters in your application configuration, e.g.:

```yaml
# config/parameters.yml
parameters:
    router.request_context.host: example.com
```

Otherwise the crawler will fall back to `localhost`, which might be wrong depending on your host configuration.

fritzmg (changes requested):

docs/dev/reference/commands.md
contaoacademy: Ok verstehe. Vielleicht dann nur den Punkt mit der Limitierung.

fritzmg:
Closing in favor of

* ba8abf3ec6835e8fe162788032ddb4f43b65a8b2
* de40dbb7da8b4cdf26a24b9f3f24cbda2b3da2c0

PR #880, opened 2021-11-22
Fix #879

### LINKED ISSUE #879: Text color of search box
On an iOS (tested on iPad, but likely also on iPhone) the text in the search box is "white on white" as the `<input>`s background is not orange. 

Adding 
```
-webkit-text-fill-color: rgb(244,124,0);
```
to `.algolia-searchbox input[type="search"]` solved that in my tests. 

I'd like to make a PR, but someone has to point me to the location of the corresponding theme file please.

fiedsch:
This fixes the Issue with white on white searchbox text on iOS devices


fritzmg (approved):

fritzmg:
Unfortunately this breaks the search box in Firefox: 

image

I will have to revert this.

PR #886, opened 2021-11-24
Update FAQ - Dateiverwaltung

fkaminski:
Frage: Kann man die Suche in der Dateiverwaltung ausblenden?

### DIFF
--- a/docs/manual/faq/_index.de.md
+++ b/docs/manual/faq/_index.de.md
@@ -124,7 +124,7 @@ Stelle außerdem sicher, dass sich keine veraltete `.htaccess` Datei im Order `/
 {{% /expand %}}
 
 {{% expand "Kann man die Suche in der Dateiverwaltung ausblenden?" %}}
-Über einen DCA Eintrag:
+Ja. Über einen DCA-Eintrag:
 
 ```yml
     //contao/dca/tl_files.php
--- a/docs/manual/faq/_index.de.md
+++ b/docs/manual/faq/_index.de.md
@@ -123,6 +123,15 @@ gekennzeichnet ist.
 Stelle außerdem sicher, dass sich keine veraltete `.htaccess` Datei im Order `/web` oder einem übergeordneten Ordner deiner Installation befindet.
 {{% /expand %}}
 
+{{% expand "Kann man die Suche in der Dateiverwaltung ausblenden?" %}}
+Über einen DCA Eintrag:
+
+```yml
+    //contao/dca/tl_files.php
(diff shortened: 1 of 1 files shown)


docs/manual/faq/_index.de.md
```diff
@@ -123,6 +123,15 @@ gekennzeichnet ist.
 Stelle außerdem sicher, dass sich keine veraltete `.htaccess` Datei im Order `/web` oder einem übergeordneten Ordner deiner Installation befindet.
 {{% /expand %}}
 
+{{% expand "Kann man die Suche in der Dateiverwaltung ausblenden?" %}}
+Über einen DCA Eintrag:
```
netzarbeiter:
```suggestion
Ja. Über einen DCA-Eintrag:
```

docs/manual/faq/_index.de.md
```diff
@@ -123,6 +123,15 @@ gekennzeichnet ist.
...
+
+```yml
```
netzarbeiter:
```suggestion
```php
```

netzarbeiter (changes requested):
Thank you, Franko

fritzmg:
I am not sure this warrants an FAQ entry? There are thousands of things you can adjust in Contao's DCA - we won't possibly list all of them in the FAQ?

Toflar:
It's still better to keep those snippets in the official documentation rather than keeping them in your personal one. But I agree, we should probably have a `Snippets` section for that?

fritzmg:
Agreed, may be a "DCA adjustments" [guide](https://docs.contao.org/manual/en/guides/) where we can list them in the same manner (but does not have to be phrased like a question then 😉). But for now I suppose we can keep it in the FAQ.

fkaminski:
> Agreed, may be a "DCA adjustments" guide where we can list them in the same manner

Maybe we could start a community call for these collections (BestOf/Awesome-List)?

fritzmg:
> Maybe we could start a community call for these collections (BestOf/Awesome-List)?

I don't think we need a community call - the page just has to be created and then over time filled up by the community :)

fkaminski:
Always looking for appropriate solutions. A consistent place in the documentation would be ideal. I assume that you already have a large number of examples for starting :-) ...

fritzmg:
> I assume that you already have a large number of examples for starting

No, not really ;). This needs to be built over time - i.e. whenever something comes up.

netzarbeiter (approved):

netzarbeiter:
https://github.com/contao/docs/commit/8989db60eace0ee1ee28c655ba33294b92faed9f

PR #890, opened 2021-12-07
Enhance initializeSystem hook description.

koertho:
Since I stumbled over the `initializeSystem` hook in the last time, I though I could add some information to the descriprion, to make it a little clearer what `before the request processing is started` means :)
What do you think?

### DIFF
--- a/docs/dev/reference/hooks/initializeSystem.md
+++ b/docs/dev/reference/hooks/initializeSystem.md
@@ -9,7 +9,7 @@ aliases:
 
 
 The `initializeSystem` hook is triggered right after the system initialization
-process is finished and before the request processing is started. This also means the request has no [contao request scope](https://docs.contao.org/dev/framework/routing/#request-scope) set and no language redirects have been executed.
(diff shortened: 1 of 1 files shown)


docs/dev/reference/hooks/initializeSystem.md
```diff
@@ -9,7 +9,7 @@ aliases:
 
 
 The `initializeSystem` hook is triggered right after the system initialization
-process is finished and before the request processing is started.
+process is finished and before the request processing is started. This also means the request has no [contao request scope](https://docs.contao.org/dev/framework/routing/#request-scope) set and no language redirects have been executed.
```
fritzmg:
```suggestion
process is finished and before the request processing is started. This also means the request has no [contao request scope](/framework/routing/#request-scope) set and no language redirects have been executed.
```

fritzmg (changes requested):
Yes, I am fine with that :). Just a nitpick.

docs/dev/reference/hooks/initializeSystem.md
koertho: Makes sense :)

fritzmg:
Thank you @koertho

PR #891, opened 2021-12-08
Expand documentation for Contao templates

fritzmg:
This PR expands on the usage of Contao templates, explaining template folders, how to set data, how to set function, etc.


docs/dev/framework/templates/_index.md
```diff
@@ -164,7 +197,7 @@ but somewhere else entirely, you could extend that template, override the `headl
 block with empty content and instead show the headline in a custom location within
 the `content` block.
 
-```php
+```html
```
fritzmg, line 200: _Note:_ this is because the syntax highlighter cannot deal with PHP code that contains multiple occurrences of `<?php … ?>`. Rendering it as `html` looks much better.

PR #893, opened 2021-12-10
Clarify cache rebuilding

fritzmg:
When clearing the cache with `--no-warmup` (which you should use), you should always also then warmup the cache afterwards - at least for the production environment. Otherwise the internal cache of Contao will be missing and thus the website will be slower.


netzarbeiter (approved):
Thannk you, Firtz

PR #900, opened 2021-12-17
Add "Web Font Loader" Link (DE/EN)

fkaminski:
See: https://community.contao.org/de/showthread.php?82105-Contao-Schriften&p=552576&viewfull=1#post552576

### DIFF
--- a/docs/manual/guides/webfont.de.md
+++ b/docs/manual/guides/webfont.de.md
@@ -143,11 +143,9 @@ deiner Contao Installation unterhalb von »files» zur Verfügung.
 Im Falle von »Google Fonts« wird dir hierzu zwar eine Download-Option angeboten, allerdings beinhaltet das jeweilige Download-
 Archiv lediglich Dateien in den Formaten ».ttf«. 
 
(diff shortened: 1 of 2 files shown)


fritzmg:
Should we mention why the new web font loader might be better? We experienced that problem a lot actually.

fkaminski:
> Should we mention why the new web font loader might be better? We experienced that problem a lot actually.

We could: 
Possibly with an info box entry with links to 
https://github.com/majodev/google-webfonts-helper/issues/98
or
https://webfontloader.altmann.de/about/#more

Maybe this would be too confusing for our documentation. Alternatively we just put the link to the new app?

What do you think?

fkaminski:
For example:

Falls du nicht sicher bist welche der genannten Webapplikationen du benutzen solltest findest du hier einen Beitrag der die Unterschiede verdeutlicht.

If you are not sure which of the mentioned web applications you should use, here is a post that clarifies the differences.

Then with link to: https://webfontloader.altmann.de/about/#more

netzarbeiter (approved):
Thank you, Franko

PR #910, opened 2021-12-21
Remove MyISAM from fallback table configuration

fritzmg:
Currently the docs suggest to switch back to `utf8` _and_ to `MyISAM` when there are problems with the row size. However, also switching to `MyISAM` might cause additional problems - and `MyISAM` will not be available in all hosting environments anymore. Thus the docs should not suggest this anymore.


netzarbeiter (approved):
Thank you, Fritz

PR #911, opened 2021-12-21
Add overview for content elements

fkaminski:
Add overview for content elements analogous to backend grouping

### DIFF
--- a/docs/manual/article-management/content-elements.de.md
+++ b/docs/manual/article-management/content-elements.de.md
@@ -14,13 +14,63 @@ den Zugriff auf einzelne Elemente bei Bedarf einschränken.
 
 ![Den Zugriff auf ein Inhaltselement einschränken](/de/article-management/images/de/den-zugriff-auf-ein-modul-einschraenken.png?classes=shadow)
 
-**Element schützen:** Das Inhaltselement ist standardmäßig unsichtbar und wird erst eingeblendet, nachdem sich ein 
-Mitglied im Frontend angemeldet hat.
 
-**Erlaubte Mitgliedergruppen:** Hier legst du fest, wer Zugriff auf das Inhaltselement hat.
+| Info |
+| -----| - |
+| **Element schützen:** | Das Inhaltselement ist standardmäßig unsichtbar und wird erst eingeblendet, nachdem sich ein Mitglied im Frontend angemeldet hat. |
+| **Erlaubte Mitgliedergruppen:** | Hier legst du fest, wer Zugriff auf das Inhaltselement hat. |
+| **Nur Gästen anzeigen:** |Das Inhaltselement ist standardmäßig sichtbar und wird ausgeblendet, sobald sich ein Mitglied im Frontend angemeldet hat. |
 
-**Nur Gästen anzeigen:** Das Inhaltselement ist standardmäßig sichtbar und wird ausgeblendet, sobald sich ein Mitglied 
-im Frontend angemeldet hat.
+
+## Übersicht
+
+{{% expand "Text-Elemente" %}}
+[Überschrift](#ueberschrift)<br>
+[Text](#text)<br>
+[HTML](#html)<br>
+[Aufzählung](#aufzaehlung)<br>
+[Tabelle](#tabelle)<br>
+[Code](#code)<br>
+[Markdown](#markdown)
+{{% /expand %}}
+
+{{% expand "Accordion" %}}
+[Einzelelement](#akkordeon)<br>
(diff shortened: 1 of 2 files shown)


docs/manual/article-management/content-elements.de.md
```diff
@@ -44,7 +94,7 @@ Das Element generiert folgenden HTML-Code:
 ```
 
 
-## Text
+## Text {#text}
```
fritzmg: Adding these wasn't really necessary, if they are the same as the default.

fkaminski: Good to know.
But we can still leave it like this - Can we?

fkaminski:
Hint:
The translation of the new content element "Custom template" it not finally translated ...

docs/manual/article-management/content-elements.de.md
```diff
@@ -1405,3 +1456,16 @@ Das Element generiert folgenden HTML-Code:
...
+
+Das Inhaltselement »Custom template« bietet die Möglichkeit beliebige Templates auszuwählen und dabei 
+eigene, individuelle Platzhalter zu definieren die im jeweiligen Template ausgegeben werden können.
+
+| Info |   |
+| ---- | - |
+| **Template data:** | Angabe eines oder mehrerer Schlüssel/Wert Paare zur Ausgabe im Template. |
+| **Inhaltselement-Template:** | Hier kannst du das Standard-Template `ce_template` überschreiben. |
+
+{{< version "4.13" >}}
```
fritzmg: Please use a spearate PR for this, so we can update the PR before 4.13 is released if necessary and release it once 4.13 is released.

fritzmg (changes requested):

docs/manual/article-management/content-elements.de.md
fkaminski: Deleting - Done.

docs/manual/article-management/content-elements.en.md
```diff
@@ -10,13 +10,65 @@ To make the creation of content as intuitive as possible, Contao provides a cont
...
+[Article](#article)<br>
+[Content element](#content-element)<br>
+[Form](#form)<br>
+[Module](#module)<br>
+[Article teaser](#article-teaser)<br>
+[Comments](#comments)
+{{% /expand %}}
+
+
+## Header {#title}
```
fritzmg:
```suggestion
## Heading {#heading}
```

docs/manual/article-management/content-elements.en.md
```diff
@@ -10,13 +10,65 @@ To make the creation of content as intuitive as possible, Contao provides a cont
...
+| ----- | - |
+| **Protect element:** | The content element is invisible by default and is only displayed after a member has logged in to the frontend. |
+| **Allowed member groups:** | Here you define who has access to the content element. |
+| **Show guests only:** | The content element is visible by default and is hidden as soon as a member has logged in to the frontend. |
+
+
+## Overview
+
+{{% expand "Text elements" %}}
+[Header](#title)<br>
```
fritzmg:
```suggestion
[Heading](#heading)<br>
```

fritzmg (changes requested):

docs/manual/article-management/content-elements.de.md
fritzmg: I don't know. It means, if you have to change a headline, you now have to change it 4 times, instead of 3 - unless you want to keep the anchor, which might be weird. In my opinion overriding the anchor should only be done when it needs to be different from the default for whatever reason. It's just a nitpick but imo we should not things that are not necessary to keep the docs and this PR as simple as possible.

netzarbeiter: Can you please remove them.

We only use them for umlauts:
## Überschrift {#ueberschrift}

fkaminski: Hm, 
linking do not work anymore (with link `[Text](#Text)`) by testing:

`## Text`
instead of 
`## Text {#text}`
???

**Working Sample:**
Link: `[Text](#text)`
Title: `## Text {#text}`

**Not Working:**
Link: `[Text](#Text)`
Title: `## Text`

fritzmg: The anchor should be `#text`, not `#Text`. Did you check the CSS ID in the HTML source?

fkaminski: Done: See last commit

netzarbeiter (approved):
Thank you, Franko

PR #913, opened 2021-12-21
Clarify XML sitemap

fritzmg:
This clarifies under which URL the XML sitemap is available since Contao 4.11. Also fixes the missing English translation.


netzarbeiter (approved):
Thank you, Fritz

PR #918, opened 2021-12-27
Update content element markdown

fkaminski:
Add "Markdown source"
Formatting "Markdown syntax" with expand layout


netzarbeiter:
Thank you, Franko.

I do not want to hide information behind expand.
It is an additional and in my eyes unnecessary click.

Besides, we had discussed on slack that we prefer to put energy into non-existing content rather than "cleaning up" existing content.

fkaminski:
This PR adds new content too. I can change the layout if you like? 
i am not in office until the end of next week and will Update as soon as possible.

netzarbeiter:
That would be great.

netzarbeiter (approved):
Thank you, Franko

PR #919, opened 2021-12-27
Add security voter example

fritzmg:
This adds a simple voter example where you can restrict admin access to the "Maintenance" back end section.


xchs (approved):

fritzmg:
Ah, this does not actually work as I intended. @aschempp what do you think would be the best practise to _expand_ voting on a certain privilige? Injecting the original voter? Decorating it?

fritzmg:
I have updated the example in af95f9e04a7256198b6d3e28ebafb13472bbaa5d - though I noticed that this works without manually setting the priority higher. @aschempp wdyt?

docs/dev/framework/security.md
```diff
@@ -163,6 +171,61 @@ $security->isGranted(ContaoNewsPermissions::USER_CAN_CREATE_ARCHIVES);
...
+        // Abstain, if we are not voting for maintenance back end module access
+        if ('maintenance' !== $subject || $attribute !== ContaoCorePermissions::USER_CAN_ACCESS_MODULE) {
+            return false;
+        }
+
+        // Get the currently logged in user
+        $user = $this->security->getUser();
+
+        // Abstain, if not back end admin
+        if (!$user instanceof BackendUser || !$user->isAdmin) {
```
Toflar: Imho that's wrong. I thought you should be checking roles? /cc @aschempp

Toflar (changes requested):

docs/dev/framework/security.md
fritzmg: If I need to access the user object anyway, then it should not matter, should it?

fritzmg:
I don't like the current approach since I have to retrieve the user in the `supports` method via the Security helper. Too bad the abstract `Voter` class does not pass the token there too. I might change it to not using the abstract class.

docs/dev/framework/security.md
fritzmg: Besides, Contao's `BackendAccessVoter` does the same:

https://github.com/contao/contao/blob/3ce5a595d30a4cb26361f03f368f9bf38faa560e/core-bundle/src/Security/Voter/BackendAccessVoter.php#L63-L67

fritzmg:
I have added an example for Contao 5 with which you can restrict editing/deleting of news articles to the original author of the news article.

fritzmg:
@aschempp one issue I keep having with implementing your own Voters for Contao permissions is that we basically have to check for everything in the `supports` method (when using the AbstractVoter) as we want to abstain if the specific conditions are not met - so that Contao's own permissions still apply. And the `supports` method unfortunately does not pass the token, which makes it a little more inconvenient to abstain based on user properties.

aschempp:
I don't think you are meant to abstain on the user token in the supports method! The supports method tells whether you support the specific attribute and subject, not any other condition of the application.

fritzmg:
But that means I cannot abstain on isAdmin for example - or other arbitrary user properties.

fritzmg:
eh, I think this is fine, merging ;)

PR #921, opened 2022-01-03
A new approach for #880 that had to be reverted

fiedsch:
Here is a new approach to solve the "text not visible in search field". I tested it with safari (iOS and MacOS) as well as Chrome and Firefox on MacOS.

### DIFF
--- a/page/static/css/theme-contao.css
+++ b/page/static/css/theme-contao.css
@@ -521,6 +521,7 @@ option {
     padding: 0 25px 0 30px;
     margin: 0;
     font-weight: 300;
+    -webkit-appearance: none;
 }
 
 .algolia-searchbox input[type="search"]::-webkit-input-placeholder {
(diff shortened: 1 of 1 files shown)


page/static/css/theme-contao.css
```diff
@@ -521,6 +521,7 @@ option {
     padding: 0 25px 0 30px;
     margin: 0;
     font-weight: 300;
+    -webkit-appearance: none;
```
fritzmg, line 525:
```suggestion
    appearance: none;
    -webkit-appearance: none;
```

Proper browsers do not need the prefix 😉 (Safari does).

fritzmg (changes requested):

fritzmg (approved):

fritzmg:
Thank you @fiedsch :)

PR #924, opened 2022-01-04
Consistently name file extension yml

fiedsch:
In most places of the docs the configuration file is called `config/config.yml` as opposed to `config/config.yaml`. For consistency reasons I would find it clearer to use the `.yml` file name extension throughout. This PR tries to fix this.

Opinions please.

### DIFF
--- a/docs/dev/framework/image-processing/image-sizes.md
+++ b/docs/dev/framework/image-processing/image-sizes.md
@@ -22,12 +22,12 @@ different settings again.
 ## Size Configuration
 
 As already mentioned you can pre-configure image sizes in the container configuration of your application 
-(`config/config.yaml`) via `contao.image.sizes` without the need to store them in the database. Here you can define 
+(`config/config.yml`) via `contao.image.sizes` without the need to store them in the database. Here you can define 
 multiple image sizes and their properties. The following is a basic example that creates two image size configurations 
 called `example` and `foobar` which will resize the images to a width of 512 or 1024 pixels respectively when selected:
 
 ```yaml
-# config/config.yaml
+# config/config.yml
 contao:
     image:
         sizes:
@@ -42,7 +42,7 @@ pixel and zoom fully into its _important part_ (if defined). Furthermore the ima
 densities and the `<img>` will have a CSS class called `example`: 
 
 ```yaml
-# config/config.yaml
+# config/config.yml
 contao:
     image:
         sizes:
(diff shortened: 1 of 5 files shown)


fritzmg (changes requested):
The official file extension for YAML is `.yaml`: https://yaml.org/faq.html
Symfony also follows this recommendation and so should Contao. I already normalised this for the dev docs in https://github.com/contao/docs/commit/31d3bfdd1336c82a049d229ab464d8b61ed82fed.

However the docs as well as Contao's source code still reference `.yml` a lot, yes.

fiedsch:
So, you mean the PR should go the other direction. Should I prepare a new one for that? Maybe including an explanation on `.yml` vs `.yaml`.

fritzmg:
Yes, the recommendation is `.yaml`. Feel free to make a suggestion :)

fiedsch:
And just to be sure: Contao—as it is based on Symfony—will also work with `.yaml` files without further changes? I never tried that …

On the other hand, there are numerous references to `.yml` files in Contao‘s language files (e.g. https://github.com/contao/contao/blob/e107ce4f1aa55c65a86d9c2256c193730e7520fb/installation-bundle/src/Resources/translations/ContaoInstallationBundle.de.xlf#L317) that would have to be changed also. Otherwise I would expect a lot of confused users.

fritzmg:
> And just to be sure: Contao—as it is based on Symfony—will also work with .yaml files without further changes? I never tried that …

Yes, both Symfony and Contao support both extensions.

> On the other hand, there are numerous references to .yml files in Contao‘s language files (e.g. https://github.com/contao/contao/blob/e107ce4f1aa55c65a86d9c2256c193730e7520fb/installation-bundle/src/Resources/translations/ContaoInstallationBundle.de.xlf#L317) that would have to be changed also. Otherwise I would expect a lot of confused users.

Yes, as I said, Contao itself still references just `.yml` in some places, which should be changed.

fiedsch:
> Yes, as I said, Contao itself still references just .yml in some places, which should be changed.

So I am closing this for now, because I don‘t think it makes sense to be ahead of the current state of Contao in the documentation.

fritzmg:
> because I don‘t think it makes sense to be ahead of the current state of Contao in the documentation.

But then your PR should stay open? ;)

fiedsch:
If we decide to go with the `.yaml` file extension my PR would be wrong. So I thought closing is appropriate.

PR #940, opened 2022-01-20
Add shortcode expand to contribute.md

fkaminski:
You were too fast with the merging :-)

I would like, for the sake of overview, also list the shortcode expand. 
(Even if I know that "someone" does not like this :-).)

Therefore: As you like ...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,17 @@ Lorem ipsum dolor sit amet ...
 ```
 
 
+### Short code expand
+
+The expand short code displays an expandable/collapsible section of text.
+
+```
+{{%expand "Expand title" %}}
+Lorem ipsum dolor sit amet ...
+{{% /expand%}}
+```
+
+
 ## New features
 
 Since we will not maintain different versions of the documentation for each minor 


fritzmg:
> You were too fast with the merging :-)

If the PR was not ready, you need to mark it as a draft.

fritzmg:
But I was just thinking: I don't think we should add any of this actually. This is already documented in the documentation of the theme. There is no need to duplicate already existing documentation and then maintain said duplication. I will adjust the `CONTRIBUTING.md` afterwards accordingly.

fkaminski:
> This is already documented in the documentation of the theme.

Certainly. But some of the Hugo theme documentation is/was implemented faster than available for our documentation (example shortcode tabs) ... 
Therefore ... ?

fritzmg:
> But some of the Hugo theme documentation is/was implemented faster than available for our documentation (example shortcode tabs) ...

Not sure what you mean?

fkaminski:
> > But some of the Hugo theme documentation is/was implemented faster than available for our documentation (example shortcode tabs) ...
> 
> Not sure what you mean?

I meant that we should mention the shortcodes that actually exist, regardless of the Hugo theme documentation. These can deviate quite temporally. This was the case with the tabs, for example. They were first available in the hugo theme and were later included in our documentation.

fritzmg:
If someone wants to use a new feature from the Learn theme, then they can open an issue so that we can update the theme accordingly. I don't see a point in duplicating the documentation from the used theme. It's just more work and maintenance with no gain.

fkaminski:
OK

Issue #606, opened 2020-11-04
Guide: Using CSS Animation within Contao


fkaminski:
A few practical examples for using and integrating CSS animations.
Maybe with examples like:

- https://github.com/animate-css/animate.css?files=1
- https://wowjs.uk/
- http://imakewebthings.com/waypoints/

Please assign to myself ...

zonky2:
extension for wowjs: https://github.com/inspiredminds/contao-wowjs

fkaminski:
I am in the process of writing the content. The main plan was a description of CSS animations (no pure SVG animations). Then in combination with wow.js or waypoint.js within Contao.  

Another option would be a description of the implementation/sample of using [Lottie](https://lottiefiles.com/) files (.json + .js player) within Contao.

Should I include this in a single file or split it into two PR's ...?

fkaminski:
My Reminder: 
Because of https://www.delac.io/WOW/#pricing
Alternative: https://github.com/aljs/wew.js with [MIT License](https://github.com/aljs/wew.js/blob/master/LICENSE.md)

fkaminski:
Contributions for the manual should be Contao specific. 
This topic would be more general. I am therefore closing this.

PR #948, opened 2022-01-25
Add Guide: Basic Animation

fkaminski:
| Info |
| --- |
| see Issue #606 |

### DIFF
--- a/docs/manual/guides/animation.de.md
+++ b/docs/manual/guides/animation.de.md
@@ -1,0 +1,236 @@
+---
+title: "Animationen"
+description: "Animationen in Contao."
+aliases:
+  - /de/anleitungen/animation/
+weight: 100
+tags: 
+  - "Theme"
+---
+
+
+Animationen beleben deine Inhalte und begegnen dir in den unterschiedlichsten Bereichen. Sparsam eingesetzt kannst du in jedem Fall 
+die gewünschte Aufmerksamkeit erzielen. Mittlerweile können Animationen in aktuellen Browsern über CSS-Angaben realisiert werden. Für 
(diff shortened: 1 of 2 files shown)


fritzmg:
I still think that things like this should not really be in the manual as it has nothing to do with Contao. The title says "Animationen in Contao" - but I do not see anything specific to Contao in that guide. The Contao Manual is the manual for... Contao. It's not supposed to be a dumping ground for anything you can do with HTML, CSS and JavaScript?

fkaminski:
Extensions like "[inspiredminds/contao-wowjs](https://extensions.contao.org/?q=wow&pages=1&p=inspiredminds%2Fcontao-wowjs)" or "[hypergalactic/contao-animate](https://extensions.contao.org/?p=hypergalaktisch%2Fcontao-animate)" contain corresponding possibilities. 

It is correct: These instructions do not contain concrete (Contao) solutions but help the general understanding. Furthermore, this reveals the wide range of possibilities with Contao.

Otherwise, some instructions would be superfluous altogether (svg/sass/webfont).

fritzmg:
> Furthermore, this reveals the wide range of possibilities with Contao.

But it's not specific to Contao? It's just what you can generally do in web development.

> Otherwise, some instructions would be superfluous altogether (svg/sass/webfont).

As I said, I don't think any of this should actually be in the Contao manual ;)

fkaminski:
As always - Only one/my suggestion. 
Ultimately your decision ...

fkaminski:
Shall we close that then? Possibly rather something for a forum post ...

netzarbeiter:
As suggested I close this pull request and would be happy about a forum post instead.

PR #961, opened 2022-01-28
Hint about CompilerPass in ExtensionPluginInterface

Toflar:
I don't know why it does not work but it doesn't. I guess @aschempp has to provide the reason here :)

### DIFF
--- a/docs/dev/framework/manager-plugin.md
+++ b/docs/dev/framework/manager-plugin.md
@@ -386,7 +386,11 @@ a certain configuration is coming from.
 {{% /notice %}}
 
 {{% notice warning %}}
-You cannot use `$container->addCompilerPass()` here because ???. See how the [respective guide](/guides/modify-container-at-compile-time/modify-container-at-compile-time/) to see how to register a compiler pass.
+You cannot use `$container->addCompilerPass()` to add a compiler pass to the app here because `getExtensionConfig()` 
+is called on every plugin during the process of merging the configuration of all the bundles (done in Symfony
+in the `MergeExtensionConfigurationPass`). If you want to register a regular compiler pass that is called **after**
+the extensions have been merged, see the [respective guide](/guides/modify-container-at-compile-time/)
+to learn more.
 {{% /notice %}}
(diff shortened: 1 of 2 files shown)


aschempp:
Can you elaborate on what _does not work_?

aschempp:
Also, adding a compiler pass in a manager plugin - especially in the interface method that is designed to override other bundle config - makes no sense. Use the bundle as usual for that, then it also works for non-managed setups.

Toflar:
I don't have a bundle. I need to make a service public in my app.

The compiler pass is never called.

Toflar:
Does not work:

```php
    public function getExtensionConfig($extensionName, array $extensionConfigs, ContainerBuilder $container)
    {
        $container->addCompilerPass(new class implements CompilerPassInterface {
            public function process(\Symfony\Component\DependencyInjection\ContainerBuilder $container)
            {
                dd('test');
            }
        });
    }
```

Works:

```php
    public function registerContainerConfiguration(LoaderInterface $loader, array $managerConfig)
    {
        $loader->load(static function (ContainerBuilder $container) {
            $container->addCompilerPass(new class implements CompilerPassInterface {
                public function process(\Symfony\Component\DependencyInjection\ContainerBuilder $container)
                {
                    dd('test');
                }
            });
        });
    }
 ```

aschempp:
`getExtensionConfig` is executed by the `MergeConfigurationPass`. I think you _could_ probably add a compiler pass that is run after merge, but adding a compiler pass in the plugin should only be done for edge cases in the app, not in any bundle. Not sure I would really recommend that here.

Toflar:
I've updated the description and extended the examples to show why it's correct and desired to have a compiler pass in your app **without** any bundle.

fritzmg (approved):
Very nice! 👍

m-vo (approved):

m-vo:
Same thing will be needed to adjust the virtual filesystem stuff in the app. Feels like I'll probably add another example to this section soon. :smile:

Issue #973, opened 2022-01-30
Add note to PHP version documentation about Contao 4.9 and PHP8 compatability


FelixOliverLange:
Hi Contao team, 

I noticed that the docs imply that Contao 4.9, as LTS release, supports PHP 8.
I attempted the upgrade from PHP 7.3 to PHP 8.0 and ran into an issue that, while it is resolvable, may not be apparent to a normal user. 

The issue comes up when attemting to use Contao 4.9.22 with PHP 8.0. While Contao itself will run, updates of e.g. Contao to 4.9.24 via the contao manager will not, as the dependency lcobucci/jwt of contao-core is locked to version 3.4.6, [which does not support PHP 8](https://packagist.org/packages/lcobucci/jwt#3.4.6). This prevents:
* Updating Contao, e.g. for security fixes
* removing other packages, e.g. due to incompatability with PHP 8 found later (in my case that was the a [cookiebar extension](https://packagist.org/packages/codefog/contao-cookiebar))

as in both cases, composer will fail stating that lcobucci/jwt does not support PHP 8. 

This can be resolved by:
* reverting back to PHP 7
* updating Contao to version 4.9.24 (as that bumps the JWT library to version 4.0.4 [which supports JWT](https://packagist.org/packages/lcobucci/jwt#4.0.4)) and removing packages not compatible with PHP 8
* switching back to PHP 8

While that's a fairly easy fix, it requires investigating said JWT library and its compatibility - and the knowledge that Contao 4.9.24 bumps this version. So maybe adding a statement to the [4.9 PHP system requirements section](https://docs.contao.org/manual/de/installation/systemvoraussetzungen/#contao-4-9-lts), explaining that the latest version of Contao 4.9 must be installed before the update, may help users. As some hosters prevent switching back to old PHP versions after an update, users may otherwise be locked into a state where they are on PHP 8 and would need to switch back temporarily but can't. As of now I don't have the feeling that the docs make this explicit, it may be an implicit assumption. 

Thanks a lot for all your work and for making Contao 4.9 compatible with PHP 8!

fritzmg:
> I noticed that the docs imply that Contao 4.9, as LTS release, supports PHP 8.

Yes, Contao 4.9 officially supports PHP 8 (see also [here](https://github.com/contao/contao/blob/4.9.24/core-bundle/composer.json#L18)).

> Contao to 4.9.24 via the contao manager will not, as the dependency lcobucci/jwt of contao-core is locked to version 3.4.6

[...]

PR #974, opened 2022-01-30
Add section about changing PHP versions

fritzmg:
See #973


Toflar (changes requested):
Can we have a reason as to why this is required? Just a simple sentence about the fact that platform requirements are also considered in Composer? Otherwise you only know that you have to do it but you don't understand why.

fritzmg:
Well, it already says

> This ensures compatibility of your installed packages with the respective PHP version.

but I'd expand it then to

> This ensures compatibility of your installed packages with the respective PHP version, since each package (including Contao itself, installed Contao extensions and other third-party packages) can require a specific PHP version that it is known to be compatible with.

wdyt?

Toflar:
👍 Maybe also "and certain PHP extensions".

docs/manual/installation/system-requirements.de.md
```diff
@@ -126,7 +126,9 @@ PHP **7.2.0** erforderlich.
 
 Falls die PHP Version einer laufenden Contao Instanz geändert werden soll, sollte immer ein volles `composer update` nach dem Wechsel
 durchgeführt werden. Dies ist besonders dann wichtig, wenn zwischen Haupt-Versionen gewechselt wird, z. B. von PHP 7.x zu PHP 8.x - oder
-umgekehrt. Dadurch wird für die installierten Pakete die Kompatibilität zu der jeweiligen PHP Version sichergestellt.
+umgekehrt. Dadurch wird für die installierten Pakete die Kompatibilität zu der jeweiligen PHP Version sichergestellt, da jedes Paket
+(inklusive Contao selbst, installierte Contao Extensions oder Dritt-Pakete) nach spezifischen PHP Versionen und PHP Extensions verlangen
```
netzarbeiter:
```suggestion
(inklusive Contao selbst, installierte Contao-Extensions oder Dritt-Pakete) nach spezifischen PHP-Versionen und PHP-Extensions verlangen
```

Toflar (approved):

netzarbeiter (changes requested):
Thank you, Fritz

netzarbeiter (approved):

FelixOliverLange:
I'm really not an expert on PHP nor on composer, so please forgive me my ignorance. But wouldn't running the update be required before the PHP version switch @fritzmg? After all, the dependencies must support the respective version before it is being run, otherwise the dependency incompatibility is noted by composer if you e.g. attempt an update or try to add an unrelated package.

fritzmg:
@FelixOliverLange composer installs the dependencies for which ever PHP version it detects. Thus running a package update before you switch will not change anything (unless you are overriding the PHP version in your `composer.json`). Also updating the dependencies to reflect a newer PHP version change _before_ you actually switch to a new PHP version will break the application even more likely, since you might then get packages that are incompatible with your _current_ PHP version.

netzarbeiter (approved):
Thank you, Fritz

PR #987, opened 2022-02-03
Fixed description of wizard reference

bennyborn:
The docs read as `wizard` just accepts a single callback but instead it behaves exactly like `load_callback` or `save_callback` thus taking an actual array of callbacks.


fritzmg (approved):

fritzmg:
Thank you @bennyborn

PR #988, opened 2022-02-05
list.sorting.paste_button callback fix function argument type hints

markocupic:


fritzmg (approved):

fritzmg:
Thank you @markocupic (sorry for the delay, went under the radar)

PR #1000, opened 2022-02-12
command line order alphabetical

netzarbeiter:


fritzmg:
I think without any weight, it would already be alphabetical, would it not?

fkaminski:
> I think without any weight, it would already be alphabetical, would it not?

But then possibly different in de/en?

fritzmg:
I think the pages are then sorted according to the filename (which is the same in de and en), but I am actually not sure anymore.

PR #1021, opened 2022-03-07
Fixed typo, formatting and description for nested insert tags and copied the content to the english manual

cliffparnitzky:


cliffparnitzky:
I've checked the insert tag from the notice info: https://github.com/contao/docs/pull/1021/files#diff-b2b1821bd572bfdfbfdb779db934d4a8a5344888dac821efdd4df28962d2e056L211

It makes no problems / page crash in my environment.

In my opinion endless loops may not be possible. Maybe @Toflar, do you have an possible scenario/explanation?

Toflar:
It is possible but as of 4.13, it will not cause an endless loop. Contao will stop on level 64.

fritzmg (approved):

netzarbeiter (approved):
Thank you, Cliff

PR #1022, opened 2022-03-20
Update theme to fix issues with Hugo >=0.94

fritzmg:
Our `menu.phtml` override is no longer necessary.


fritzmg:
@sheeep should work now with Hugo >=0.94

PR #1031, opened 2022-04-09
Update dca.md

raum51:
adapted onload_callback in full example of tl_parts.php

The example code for onload_callback of tl_parts.php differed in section tl_parts > config and full example.

### DIFF
--- a/docs/dev/guides/dca.md
+++ b/docs/dev/guides/dca.md
@@ -506,8 +506,11 @@ $GLOBALS['TL_DCA']['tl_parts'] = [
         ],
         'onload_callback' => [
             function () {
-                $db = Database::getInstance();
-                $pid = Input::get('pid');
+                $db = \Contao\Database::getInstance();
+                $pid = \Contao\Input::get('pid');
+                if (empty($pid)) {
+                    return;
+                }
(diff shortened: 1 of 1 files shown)


fiedsch (commented):
This should not be required as the example also has `use Contao\Database` (at the [top of the example](https://github.com/contao/docs/blob/347051eb72559160e9bd9961651728fbc03e1134/docs/dev/guides/dca.md?plain=1#L493)) 

The usage of `use ...` and "inline class paths" might be unified though, because it is mixed throughout the examples.

raum51:
So the better way would be to make the same usage of `use ...` in the parts > config section as in the full example (and adopt `$db = ...` and `$pid = ...`), BUT in the full example following code is missing:
```
                if (empty($pid)) {
                    return;
                }
```

fiedsch:
I would go with using `use ...` in all places. @fritzmg or @netzarbeiter will comment on what they would recommend.

The missing `$pid` part is another issue (which I did not check).

fritzmg (approved):
Originally there was no use statement at all. It's just personal preference.

fritzmg:
Thank you @raum51

PR #1032, opened 2022-04-09
Fix "In your application" example for the virtual filesystem

m-vo:
You cannot modify the container directly, as the core's services won't be in there at that point. Registering a compiler pass as an anonymous class works, though.


fritzmg (approved):

fritzmg:
Thank you @m-vo

PR #1034, opened 2022-04-20
Make the access key documentation browser-aware

aschempp:
I was annoyed that our accesskey documentation only works on Windows. So I built a small JS library that can toggle HTML tags based on browser or platform, and also added a shortcode for access keys. With this change, the docs will always show the correct access keys for the current browser (based on the Mozilla docs) 😎 

I'm not sure if there are more access keys in the docs, but they could easily be updated later


netzarbeiter (approved):
Thank you, Andy

PR #1039, opened 2022-04-29
add new guide - displaying own website in iframe of another domain

MDevster:


MDevster:
see https://github.com/contao/docs/issues/177

fritzmg:
Hm, this still won't override the `X-Frame-Options` response header and thus still not allow the Contao site to be used as an iframe on a foreign domain, will it?

netzarbeiter:
Due to inactivity, I am closing this pull request.

PR #1042, opened 2022-05-16
Replace absolute flag with attribute

ausi:
See https://github.com/contao/contao/pull/3146

### DIFF
--- a/docs/manual/article-management/insert-tags.de.md
+++ b/docs/manual/article-management/insert-tags.de.md
@@ -204,7 +204,7 @@ Insert-Tags, die als Ausgabe eine ID oder Alias haben, können grundsätzlich ve
 
 | Insert-Tag | Ausgabe |
 |:-----------|:--------|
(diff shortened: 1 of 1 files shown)


Toflar (approved):

Toflar:
Ah, maybe I was too quick merging this! This doesn't work in 4.9, does it?
The `::absolute` is actually mentioned further down. And if we want this adjustment, we also need it in English.

ausi:
> This doesn't work in 4.9, does it?

Yes, sorry 🙈 Was changed in 4.12: https://github.com/contao/contao/commit/36dc94777fd1da8a91eed022804869f0f898fd49

Will do a folloup PR

fritzmg:
@Toflar please let @netzarbeiter handle merges into the manual, as he also takes care of it getting translated in case the translation was not provided.

Toflar:
I did check for the translation, don't worry :)

PR #1045, opened 2022-05-25
Update settings.de.md

fkaminski:
Erweiterter Hinweise bez. Mail-Transport Settings und Encoding

### DIFF
--- a/docs/manual/system/settings.de.md
+++ b/docs/manual/system/settings.de.md
@@ -600,7 +600,7 @@ dazu auch die Informationen in der offiziellen [Symfony Dokumentation][SymfonyMa
 Falls der Benutzername oder das Passwort Sonderzeichen verwendet, müssen diese "URL enkodiert" werden. Es gibt
 verschiedene Online-Services, mit denen man auf einfache Weise eine beliebige Zeichenfolgen URL-encoden kann, z. B.
 [urlencoder.org](https://www.urlencoder.org/). Enkodiere den Benutzernamen und das Passwort separat, nicht gemeinsam
-mit dem Doppelpunkt. Die jeweiliege Kodierung eines Sonderzeichen muß mit % eingeleitet werden: Also z.B. %%23 für das Sonderzeichen #.
+mit dem Doppelpunkt. Bei Verwendung der `config.yml` muß die jeweiliege Kodierung eines Sonderzeichen mit % eingeleitet werden: Also z.B. %%23 für das Sonderzeichen #.
 {{% /notice %}}
 
 {{% notice tip %}}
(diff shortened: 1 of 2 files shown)


docs/manual/system/settings.de.md
```diff
@@ -600,7 +600,7 @@ dazu auch die Informationen in der offiziellen [Symfony Dokumentation][SymfonyMa
 Falls der Benutzername oder das Passwort Sonderzeichen verwendet, müssen diese "URL enkodiert" werden. Es gibt
 verschiedene Online-Services, mit denen man auf einfache Weise eine beliebige Zeichenfolgen URL-encoden kann, z. B.
 [urlencoder.org](https://www.urlencoder.org/). Enkodiere den Benutzernamen und das Passwort separat, nicht gemeinsam
-mit dem Doppelpunkt.
+mit dem Doppelpunkt. Die jeweiliege Kodierung eines Sonderzeichen muß mit % eingeleitet werden: Also z.B. %%23 für das Sonderzeichen #.
```
fritzmg: That depends on where you use it. A `%` character only needs to be doubled in your YAML config - but not if you use environment variables (via your `.env.local` for example). See https://symfony.com/doc/4.4/configuration.html#configuration-parameters

fritzmg (changes requested):

docs/manual/system/settings.de.md
fkaminski: However, should this not already be explicitly distinguished so?

fkaminski: Add a hint for config.yml usage

fritzmg: > However, should this not already be explicitly distinguished so?

I don't know what you mean? You need to distinguish that in your sentence that you want to add ;). e.g.:

> Note that in your YAML configuration files any `%` character [needs to be escaped](https://symfony.com/doc/4.4/configuration.html#configuration-parameters) with another `%`, so `%%23` for example for the the special character `#`. This will not be necessary if you defined your mailer DSN in an environment variable and you are only referencing that variable in your YAML configuration.

docs/manual/system/settings.de.md
```diff
@@ -600,7 +600,7 @@ dazu auch die Informationen in der offiziellen [Symfony Dokumentation][SymfonyMa
 Falls der Benutzername oder das Passwort Sonderzeichen verwendet, müssen diese "URL enkodiert" werden. Es gibt
 verschiedene Online-Services, mit denen man auf einfache Weise eine beliebige Zeichenfolgen URL-encoden kann, z. B.
 [urlencoder.org](https://www.urlencoder.org/). Enkodiere den Benutzernamen und das Passwort separat, nicht gemeinsam
-mit dem Doppelpunkt.
+mit dem Doppelpunkt. Bei Verwendung der `config.yml` muß die jeweiliege Kodierung eines Sonderzeichen mit % eingeleitet werden: Also z.B. %%23 für das Sonderzeichen #.
```
fritzmg:
```suggestion
mit dem Doppelpunkt. Bei Verwendung der `config.yaml` muß die jeweilige Kodierung eines Sonderzeichen mit `%` eingeleitet werden: Also z.B. `%%23` für das Sonderzeichen `#`.
```

fritzmg (changes requested):

docs/manual/system/settings.de.md
fkaminski: Thank you

fritzmg (approved):

netzarbeiter (approved):
Thank you, Franko

PR #1048, opened 2022-06-24
Hooks EventListener SetNewPassword $password is encrypted

d0p3fish:


fritzmg:
If the password really is passed encrypted - are you sure this isn't actually a bug in Contao?

fritzmg:
Yeah, looks like this has always been the encrypted password, even in Contao 2.11.

fritzmg:
Thank you @d0p3fish

PR #1051, opened 2022-06-29
Add information about "Click to edit"

ameotoko:
This documents the "Click to edit" feature in backend. The text is taken directly from [release notes](https://contao.org/en/news/contao-3_1_RC1.html).


ameotoko:
If https://github.com/contao/contao/pull/4904 gets merged, the behaviour of `ctrl-click` and `ctrl-shift-click` will be the opposite, starting from `5.0`

May be we should wait and see, how that plays out.

ameotoko:
https://github.com/contao/contao/pull/4904 is merged, so I guess I should either

1. document the new behavior as default and mention that it worked the other way around prior to `5.x`
2. leave as it is and mention that it will be reversed in `5.x`

Suggestions?

fritzmg:
Seems like there is some disagreement: https://github.com/contao/contao/pull/4910

ameotoko:
Contao 5 is released, what shall we do with this now?

netzarbeiter (approved):
Thank you, ameotoko

PR #1061, opened 2022-07-26
Fixed typo

Timo-Schroeder:
I think you switched some words around here. This should be a better formulation.


netzarbeiter (approved):
Thank you, Timo

PR #1063, opened 2022-08-09
Remove unnecessary translation

ameotoko:
I think it was not supposed to be translated; otherwise it looks strange – why English page title produces German alias😁


fritzmg (approved):

netzarbeiter (approved):
Thank you, ameotoko

PR #1064, opened 2022-08-11
Add Info about the MAILER_DSN environment variable

fkaminski:
See [#5063](https://github.com/contao/contao/pull/5063)


fritzmg (changes requested):
Actually we should just turn it around, because all actively supported Contao versions already support `MAILER_DSN` anyway (including 4.9 for example). Instead you could add a notice of type `info` that this variable was called `MAILER_URL` previously and since Contao 5.0 only `MAILER_DSN` will be supported.

However, I do notice another issue: for Contao 4.13 and above we need to refer to the Symfony Mailer documentation. That would be another notice.

fritzmg (approved):

PR #1069, opened 2022-08-23
Describe how to maintain Many-To-Many or One-To-Many relations with the DCA picker

richardhj:
I am using the DCA picker on Doctrine entities and this is how I got it working (real Doctrine associations instead of a serialized string in the database).

### DIFF
--- a/docs/dev/reference/widgets/picker.md
+++ b/docs/dev/reference/widgets/picker.md
@@ -131,6 +131,7 @@ use App\Entity\ContentElement;
 use App\Repository\ArticleRepository;
 use Contao\CoreBundle\DependencyInjection\Attribute\AsCallback;
 use Contao\DataContainer;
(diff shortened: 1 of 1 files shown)


fritzmg (changes requested):
I am not exactly sure I can merge this as is. I think the example will not be clear to the majority, as you are using entities for content elements and articles (if I understand this correctly)? In order to be of use for the rest it should use Contao's models instead, otherwise it is too specific to your personal Contao instance.

richardhj:
> I think the example will not be clear to the majority, as you are using entities for content elements and articles (if I understand this correctly)? In order to be of use for the rest it should use Contao's models instead, otherwise it is too specific to your personal Contao instance.

In reality, I am using entities like Product and ProductCategory but adapted it to Article/ContentElement to match the docs entry :)

PR #1070, opened 2022-08-25
Update info on Twig namespaces

ameotoko:
This updates Twig Namespaces section with more precise information about which bundle templates directory is registered automatically, and which can optionally be added.

### DIFF
--- a/docs/dev/framework/templates/architecture.md
+++ b/docs/dev/framework/templates/architecture.md
@@ -92,10 +92,44 @@ worry, you don't have to remember this):
 
 | Directory | Namespace | Order
 |-|-|-|
-| Any bundle's template directory:<br>`/vendor/…/templates`<br>`/vendor/foo/bar/contao/templates` | `@Contao_<bundle>`<br>`@Contao_FooBarBundle` | 1 |
+| Any bundle's template directory:<br>`/vendor/foo/bar/src/Resources/contao/templates` | `@Contao_FooBarBundle` | 1 |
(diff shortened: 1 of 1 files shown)


fritzmg:
In Symfony 6 the default is different. May be we should link to the Symfony Docs instead?

ameotoko:
What do you mean? The [section on namespaces](https://symfony.com/doc/current/templates.html#template-namespaces) is identical between `5.4` and `6.1` branches.

fritzmg:
I mean https://symfony.com/doc/current/bundles/best_practices.html#directory-structure

In Symfony 6 you can extend from `AbstractBundle` which returns the root directory of the package in `getPath` by default.

ameotoko:
Will the [example on creating a bundle](https://docs.contao.org/dev/getting-started/extension/#creating-the-bundle) be updated as well?

fritzmg:
Yes, I definitely want to update that as well (though not by using `AbstractBundle`).

ameotoko:
How about now, @fritzmg?

fritzmg (approved):

fritzmg:
Unfortunately there is a conflict with this PR now. The section has moved to https://github.com/contao/docs/blob/main/docs/dev/framework/templates/architecture.md

ameotoko:
I rebased and brought my edits to the new location "as is". However, I was not following the core team's work in past months, so I may not be on top of the things with this PR anymore.

fritzmg (approved):

fritzmg:
Thank you @ameotoko

PR #1071, opened 2022-08-26
Update Deployer guide

ameotoko:
Small updates to the article about using Deployer.

 * There's no need to reference an RC version, since Deployer 7 has been [released](https://github.com/deployphp/deployer/releases/tag/v7.0.0) (which also contains a [bug fix](https://github.com/deployphp/deployer/pull/3200) to Contao recipe).

 * An example on excluding folders from pre-set `shared_dirs` has been added. Not everybody uses Contao Manager and `parameters.yml`, but Contao recipe deploys them by default as empty directories.

 * I find the wording about Contao "slowly migrating towards using `public/` instead of `web/`" a little bit unfortunate. The guide claims to be intended for Contao 4.13 and up, yet `public/` is already a default since 4.12

### DIFF
--- a/docs/manual/guides/deployer.de.md
+++ b/docs/manual/guides/deployer.de.md
@@ -21,7 +21,7 @@ Zuerst musst du Deployer installieren: [https://deployer.org/docs/][1]
 Du kannst Deployer entweder global oder per Projekt (mit Composer) installieren. Der Kommandozeilenbefehl lautet
 entsprechend `dep` oder `./vendor/bin/dep`.
 
-Bevor du weitermachst, stelle sicher, dass du mindestens Version _7.0.0-rc.5_ installiert hast (`dep --version`).
+Bevor du weitermachst, stelle sicher, dass du mindestens Version _7.0_ installiert hast (`dep --version`).
 
 Jetzt kannst du deine `deploy.php`-Datei in dem Projekt erstellen:
 
--- a/docs/manual/guides/deployer.de.md
+++ b/docs/manual/guides/deployer.de.md
@@ -178,14 +178,6 @@ before('deploy:publish', 'contao:manager:download');
 after('contao:manager:download', 'contao:manager:lock');
 ```
 
-Verwendest du Contao Manager nicht, dann kannst du den leeren Ordner von "Shared folders" ausschließen:
-
-```php
-// deploy.php
-
-set('shared_dirs', array_diff(get('shared_dirs'), ['contao-manager']));
-```
-
 
 ### Probleme mit dem Symlink und OPCache / APCu
 
--- a/docs/manual/guides/deployer.de.md
+++ b/docs/manual/guides/deployer.de.md
@@ -18,10 +18,11 @@ Diese Anleitung bezieht sich auf die Deployer-Version >=7.0 und Contao-Version >
 
 Zuerst musst du Deployer installieren: [https://deployer.org/docs/][1]
 
-Du kannst Deployer entweder global oder per Projekt (mit Composer) installieren. Der Kommandozeilenbefehl lautet
-entsprechend `dep` oder `./vendor/bin/dep`.
+```bash
+composer require --dev deployer/deployer
+```
 
-Bevor du weitermachst, stelle sicher, dass du mindestens Version _7.0_ installiert hast (`dep --version`).
+Bevor du weitermachst, stelle sicher, dass du mindestens Version _7.0_ installiert hast (`vendor/bin/dep --version`).
 
 Jetzt kannst du deine `deploy.php`-Datei in dem Projekt erstellen:
 
@@ -148,7 +149,7 @@ before('deploy', 'encore:compile');
 
 ## Und nun: Deployen!
 
-Nachdem wir alles konfiguriert haben, können wir jetzt `dep deploy` ausführen und das Projekt auf den Webserver
+Nachdem wir alles konfiguriert haben, können wir jetzt `vendor/bin/dep deploy` ausführen und das Projekt auf den Webserver
 deployen.
 
 
--- a/docs/manual/guides/deployer.de.md
+++ b/docs/manual/guides/deployer.de.md
@@ -113,8 +113,9 @@ hinter Deployer ist, dass es keine Downtime bei Updates gibt. Deswegen werden ro
 Document-Root von deinem vHost musst du entsprechend auf `/current/public` (bzw. `/current/web`) einstellen. Ein
 komplettes Beispiel für einen Document-Root wäre: `/var/www/foobar/html/example.org/current/public`.
 
-Contao migriert langsam zum `/public`-Ordner als Web-Root. Wenn du in Contao 4.13 noch den `/web`-Ordner verwendest,
-dann definiere diesen entsprechend, damit Deployer das auch weiß.
+{{% notice "info" %}}
+Standardmäßig verwendet Contao `/public`-Ordner als Web-Root. Wenn deine Contao-Installation noch den Legacy `/web`-Ordner
+verwendet, dann definiere diesen entsprechend in `composer.json`, damit Deployer das auch weiß.
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


docs/manual/guides/deployer.en.md
```diff
@@ -19,7 +19,7 @@ If not done yet, install Deployer as described here: [https://deployer.org/docs/
 
 You can either install Deployer globally or per project and use the command `dep` or `./vendor/bin/dep` respectively.
```
richardhj: The global install is not recommended anymore. Instead, it is recommended to install Deployer per project and create a bash alias (this changed in rc.6 or so).

See https://deployer.org/docs/7.x/installation

Can you updated this as well?

docs/manual/guides/deployer.de.md
```diff
@@ -113,8 +113,9 @@ hinter Deployer ist, dass es keine Downtime bei Updates gibt. Deswegen werden ro
 Document-Root von deinem vHost musst du entsprechend auf `/current/public` (bzw. `/current/web`) einstellen. Ein
 komplettes Beispiel für einen Document-Root wäre: `/var/www/foobar/html/example.org/current/public`.
 
-Contao migriert langsam zum `/public`-Ordner als Web-Root. Wenn du in Contao 4.13 noch den `/web`-Ordner verwendest,
-dann definiere diesen entsprechend, damit Deployer das auch weiß.
+{{% notice "info" %}}
+Standardmäßig verwendet Contao `/public`-Ordner als Web-Root. Wenn deine Contao-Installation noch den Legacy `/web`-Ordner
```
richardhj:
```suggestion
Contao verwendet standardmäßig den `/public`-Ordner als Web-Root. Wenn deine Contao-Installation noch den alten `/web`-Ordner
```

docs/manual/guides/deployer.de.md
```diff
@@ -176,6 +177,14 @@ before('deploy:publish', 'contao:manager:download');
 after('contao:manager:download', 'contao:manager:lock');
 ```
 
+Verwendest du Contao Manager nicht, dann kannst du den leeren Ordner von "Shared folders" ausschließen:
```
richardhj: Even if the Contao manager is not used, sharing the folder does not hurt. I'd rather skip this info :-)

docs/manual/guides/deployer.en.md
```diff
@@ -168,6 +170,14 @@ before('deploy:publish', 'contao:manager:download');
 after('contao:manager:download', 'contao:manager:lock');
 ```
 
+If you don't use Contao Manager, you can prevent empty folder from being deployed by excluding it from shared folders:
```
richardhj: see above.

docs/manual/guides/deployer.de.md
ameotoko: Sure, doesn't hurt, but I don't like the idea of extra empty folders hanging around in my production servers :) Contao does not come with Manager by default, so why should I have empty `contao-manager` folder?

docs/manual/guides/deployer.en.md
ameotoko: Done in [93eaf6f](https://github.com/contao/docs/pull/1071/commits/93eaf6f232b1063cbf0e07b6166fed9f889f8df5)

docs/manual/guides/deployer.de.md
```diff
@@ -123,7 +125,7 @@ dann definiere diesen entsprechend, damit Deployer das auch weiß.
   }
 }
 ```
-
+{{% /notice %}}
```
fritzmg, line 128: This would not work for Contao 4.9. I think this should be reverted?

docs/manual/guides/deployer.de.md
fritzmg: I agree with @richardhj. It would be weird for this guide to use the Contao recipe - only to then remove things that the recipe already provides. Any advanced user will customize this anyway.

fritzmg (changes requested):

docs/manual/guides/deployer.de.md
ameotoko: The guide is intended for Contao 4.13 and up though?

docs/manual/guides/deployer.de.md
ameotoko: Ok, I'll remove this.

docs/manual/guides/deployer.de.md
richardhj: > The guide is intended for Contao 4.13 and up though?

The reason for this is the usage of the migrate and backup commands in the recipe.

docs/manual/guides/deployer.de.md
ameotoko: > It would be weird for this guide to use the Contao recipe - only to then remove things that the recipe already provides.

I feel weird that the recipe creates unused files and folders😁 but that question does not belong in this repository, I agree.

docs/manual/guides/deployer.de.md
fritzmg: `contao:migrate` is already available in Contao 4.9.

docs/manual/guides/deployer.de.md
fritzmg: That's just how recipes work. I do not use any recipes for my deployer Scripts (except `recipe/common.php`) as they all do things that I do not want ;)

docs/manual/guides/deployer.de.md
ameotoko: @richardhj I don't see usages of `contao:backup:*` in the recipe. Can I then safely remove the notice on top (about 4.13), and revert/update this change?

richardhj: Ah, no, it is about the maintenance tasks. Added a in Contao 4.13. See lines 70ff.

ameotoko: So that means Contao 4.9 is irrelevant here @fritzmg.

Other (an advanced) option would be updating the recipe to detect Contao version and/or presence of `maintenance-mode` or Lexik bundle, and execute (or skip) the command accordingly.

fritzmg: The guide only mentions `contao:maintenance:disable`. `contao:maintenance:enable` isn't even used. I think this line can be removed. `contao:maintenance:enable` is (mostly) not necessary anyway as the point of using Deployer is also zero down time deployments. (Using `contao:maintenance:enable` would negate that.)

ameotoko: > `contao:maintenance:enable` isn't even used

It is invoked by the recipe. `contao:maintenance:disable` is mentioned in context of failed deployment, when prod installation would be left hanging in maintenance-enabled state.

richardhj: > The guide only mentions `contao:maintenance:disable`. `contao:maintenance:enable` isn't even used. I think this line can be removed. `contao:maintenance:enable` is (mostly) not necessary anyway as the point of using Deployer is also zero down time deployments. (Using `contao:maintenance:enable` would negate that.)

However, the database is shared, so we must make sure no invalid transactions while migrating.

ameotoko: Yes, it is meant (as far as I understand) as a mean of locking the database while `contao:migrate` is performed.

fritzmg: I see. But then we should remove any mentions of zero down time in the guide (as there is now down time involved when using this recipe).

richardhj (approved):

docs/manual/guides/deployer.de.md
netzarbeiter: Can I merge the pull request if I adjust the downtime text as follows?

_Die Idee hinter Deployer ist, dass es eine möglichst kurze Downtime bei Updates gibt._

_The idea behind Deployer is that there is the shortest possible downtime during updates._

richardhj: Yes, great suggestion! :)

netzarbeiter (approved):
Thank you, ameotoko

PR #1084, opened 2022-10-06
Add Twig inheritance page

fkaminski:
DE only (EN Translation after review).
Respects hugo weight variable in relation to PR #1082
(Nothing new here. Rather analogous to the PHP template navigation structure)


fkaminski:
@m-vo Should we add the hierarchy examples here?

PR #1087, opened 2022-10-14
List core twig function and filter

fkaminski:
As discussed in slack: list of contao core twig functions, filter (macros, tags etc.) ...

### DIFF
--- a/docs/dev/framework/templates/twig.md
+++ b/docs/dev/framework/templates/twig.md
@@ -591,6 +591,23 @@ Building and outputting HTML attributes in a structured way:
 {# Outputs: <div data-foo="bar">Demo</div> #}
 ```
 
+This function includes the following methods:
+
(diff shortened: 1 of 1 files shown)


fkaminski:
Maybe we should list existing macros here too.
For example from »[image studio](https://github.com/contao/contao/blob/5.x/core-bundle/templates/Image/Studio/_macros.html.twig)« (concerning [#5367](https://github.com/contao/contao/issues/5367))

fkaminski:
We should probably also consider this: [#4657](https://github.com/contao/contao/pull/4657)
(Allow adding document scoped content from within Twig templates)

fkaminski:
Closing as discussed in contao-call.
Possibly then add here:
https://docs.contao.org/dev/framework/templates/creating-templates/#template-features

PR #1092, opened 2022-11-01
Add ext-fileinfo to system requirements

fritzmg:
Through `league/flysystem` (which requires `league/mime-type-detection`) Contao 4.13 has a new system requirement for the `fileinfo` PHP extension.


netzarbeiter (approved):
Thank you, Fritz.

PR #1100, opened 2022-11-21
Update manager-plugin.md

Tastaturberuf:


fritzmg (changes requested):
This is not correct, the Contao Manager Plugin only loads the following classes automatically:

* `\App\ContaoManager\Plugin`
* `\ContaoManagerPlugin`

I think you are confusing FQCNs with file paths.

Tastaturberuf:
Ah yes, i thats FQCNs. Thank you @fritzmg

Issue #840, opened 2021-09-21
Document new config for the list of allowed protocols


MarkejN:
With https://github.com/contao/contao/pull/3488 the following config is going to be added:
```yaml
contao:
    sanitizer:
        allowed_url_protocols:

            # Defaults:
            - https
            - http
            - ftp
            - mailto
            - tel
            - data
            - skype
            - whatsapp
```
Versions: probably 4.9.25+ and 4.13+

MarkejN:
Where should I add this? The large config block doesn't allow any version notes, correct?

netzarbeiter:
I would add it [here](https://docs.contao.org/manual/de/system/einstellungen/#config-yml).

Toflar:
I think you can use a comment in the example and say `# (as of Contao 4.13)` or something similar :)

MarkejN:
https://github.com/contao/docs/pull/1101

PR #1101, opened 2022-11-23
Add contao.sanitizer config

MarkejN:
See https://github.com/contao/docs/issues/840

The config is available since 4.9.25, 4.12.6 and 4.13. I don't think this info is needed here, right?

I can add the comment to the ConfigTreeBuilder afterwards.

### DIFF
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -174,6 +174,20 @@ contao:
 
         # Allows to configure the default HttpClient options (useful for proxy settings, SSL certificate validation and more).
         default_http_client_options: []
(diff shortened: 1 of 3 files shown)


fritzmg:
> The config is available since 4.9.25, 4.12.6 and 4.13. I don't think this info is needed here, right?

Indeed :).

Btw. the whole config tree is just a copy & paste from the `vendor/bin/contao-console config:dump-reference contao` output.

fritzmg (approved):

fritzmg:
Thank you @MarkejN

PR #1120, opened 2023-01-02
Fix custom controller example

Toflar:
The current example does not work at all. I don't know how this could make it in the docs. `git blame` indicates that this was part of the `Initial commit` by @netzarbeiter 4 years ago so I suppose it was somehow taken over from somewhere.

However, there has never been a `@ContaoCore/Backend/be_page.html.twig` template so this whole example never worked at all.
While trying to get this to work, I noticed that @m-vo considerably improved this use case in https://github.com/contao/contao/pull/3180. 

I thus decided to adjust this example to 4.13 only. Again, previously it didn't work at all and I don't feel like trying to get something to work in 4.9 just for the documentation's sake.


fritzmg:
> However, there has never been a `@ContaoCore/Backend/be_page.html.twig` template so this whole example never worked at all.

It did, it was introduced in Contao 4.4: https://github.com/contao/core-bundle/pull/512

However, it was removed in Contao 5.0.

fritzmg:
I'd like to keep compatibility with Contao 4.13 here.

However, it's a bummer that the template won't work in Contao 5.0 anymore. @m-vo any way we can add `be_page.html.twig` back somehow?

Otherwise we can just introduce tabs for the template (and may be controller, not sure which version introduced the backend route prefix)

m-vo:
I'll look into it. I have some ideas for improving the custom back end controller situation anyways but they are not ready, yet.

docs/dev/guides/back-end-routes.md
```diff
@@ -88,31 +79,17 @@ Our route will render the template `my_backend_route.html.twig` which must be pl
...
-
-{% block headline %}
-    Not only the content of the `title`-tag but also the title of the content section.
-{% endblock %}
+{% extends "@Contao/be_main" %}
 
-{% block error %}
-    Will be placed within the error block.
-{% endblock %}
+{% block main_headline %}Headline{% endblock %}
```
rabauss: There is no block `main_headline` - you need to set the `headline` via template parameter like `title`!

fritzmg:
This does not seem to work at all in Contao 4.13 either. @m-vo we really need to fix this, otherwise it's impossible to create a back end controller that is compatible with Contao 4.13 and 5.

rabauss:
For me it's working in my extension - I had to do some changes - see: https://github.com/bwein-net/contao-database-backup/commit/eae5049f828676f0fe77a9e39cd3bdfaf121b910

fritzmg:
Ah yes, thx, nevermind, I did not extend from `AbstractBackendController`.

fritzmg:
Thank you @Toflar !

PR #1122, opened 2023-01-09
Added a simplified routing plugin example

Toflar:
I can't remember myself by heart. I always have to copy so I figured, I'll add it to the docs ;)

### DIFF
--- a/docs/dev/framework/manager-plugin.md
+++ b/docs/dev/framework/manager-plugin.md
@@ -566,6 +566,27 @@ class Plugin implements RoutingPluginInterface
 }
 ```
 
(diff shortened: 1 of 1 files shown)


fritzmg (approved):
I didn't even know about this :)

Btw. side note: within Contao we use `$resolver->resolve(__DIR__.'…')` rather than `$resolver->resolve('@BundleName…')`. What's the current best practice? Should we change it in the docs to using `__DIR__`?

Toflar:
I still use the bundle notation - both work :)

fritzmg:
Thank you @Toflar :)

PR #1124, opened 2023-01-11
Fix syntax highlight in custom environment variable section

JanMalte:


docs/manual/system/settings.en.md
```diff
@@ -544,12 +544,12 @@ of the `config/parameters.yaml` and thus you should not use these parameters, if
 
 The following example shows how to define the system administrator's e-mail address using a custom environment variable in the `.env` file and referencing it in the `config.yml` file.
 
-`yaml
+```dotenv
```
fritzmg:
```suggestion
```yaml
```
The usage of `yaml` was deliberate, to have some syntax highlighting (so that comments are coloured differently from the rest).

fritzmg (changes requested):

docs/manual/system/settings.en.md
JanMalte: The `.env`  file is not a YAML file and `dotenv` type should highlight comments as well.

fritzmg: As far as I checked, there is no `dotenv` type. No syntax highlighting will take place if you use `dotenv`.

JanMalte, line 547:
```suggestion
```ini
```

@fritzmg I would suggest to use `ini` instead. This is supported and has more or less the same valid syntax.
grafik

JanMalte:
```suggestion
```ini
```

@fritzmg I would suggest to use `ini` instead. This is supported and has more or less the same valid syntax.
grafik

fritzmg: Yes, that's fine with me :). Please also adjust the German version of the page accordingly.

JanMalte: @fritzmg Thank you very much for your feedback and review of the changes.

fritzmg (approved):

fritzmg:
Thank you @JanMalte

PR #1125, opened 2023-01-12
Update description for checkbox wizard

koertho:
Since this widget is already documented, we should not display "Not yet documented" there :)


Toflar (approved):

fritzmg (approved):

fritzmg:
Thank you @koertho

PR #1131, opened 2023-01-19
Twig Docs

m-vo:
Currently a work in progress with some open todos and some ~~missing~~ PRs in the Contao core: https://github.com/contao/contao/pull/5619, https://github.com/contao/contao/pull/5717

 We also might want to discuss some things regarding best practices before releasing it.

That said: enjoy reading. :slightly_smiling_face: As it is quite a lot of text and formatting, I suggest checking out the pull request and building the docs locally for that.

@fritzmg I created two new shortcodes (`example` and `best-practice`) because I needed to keep examples together and wanted to be able to highlight best practices without putting them all in a `notice`. Is that ok? I hope I did implement them correctly.


page/layouts/shortcodes/example.html
```diff
@@ -0,0 +1,11 @@
+{{ $_hugo_config := `{ "version": 1 }` }}
```
m-vo: Not sure what that does. It is part of some of the other shortcodes…

fritzmg: Only `taxonomylist` has that, no? I think that came from an example, presumably to influence some parsing behavior. I think you can remove that (and test whether the output is as expected).

m-vo: Seems unnecessary. :smile: I removed it in 0d29ba97d5cd9052293772be9fbd52ef3517af0d.

m-vo:
This is ready for a first round of reviews.

docs/dev/framework/templates/getting-started.md
```diff
@@ -0,0 +1,155 @@
+---
+title: "1. Getting started"
```
fritzmg:
```suggestion
title: "1. Getting started"
```

The numbering is not required and we do not use that anywhere else. You can use the `weight` property, if you want to order the articles in a specific way.

docs/dev/framework/templates/getting-started.md
```diff
@@ -0,0 +1,155 @@
...
+But ultimately, the biggest selling point for us is security. Twig embraces *output encoding*, something that we
+desperately try to move to for a long time. That is, why Twig has been made a core requirement and will replace our old
+existing template system and also why you should familiarize yourself with it.
+
+## Learning the syntax
+Twig templates have their own syntax, but don't be afraid, you'll quickly find your way. Here, is an example of a PHP
+template, that was translated into an equivalent Twig template. (In case you're new to Contao, you might want to directly
+focus on the right side and ignore the old stuff on the left. 😉)
+
+<div style="display:grid;grid-template-columns:1fr 40px 1fr;margin:-20px 0">
```
fritzmg: I think we should introduce a CSS class here - so that we can also make it responsive (will get squished too much on mobile 🙃).

fritzmg (commented):
> @fritzmg I created two new shortcodes (example and best-practice) because I needed to keep examples together and wanted to be able to highlight best practices without putting them all in a notice. Is that ok? I hope I did implement them correctly.

I think I am fine with the `example` shortcode, however I am not sure about the `best-practice` one. I think all these instances should just be a `tip` notice (and may be start with **Best practice:**).

Also please add an empty line after the headlines 🙂

docs/dev/framework/templates/getting-started.md
m-vo: Thanks for pointing out the `weight` property. Makes sense. :+1: 

Adjusted in 3db3be5fd8eaaedca8d7580ba53e66326d9bf79f.

docs/dev/framework/templates/getting-started.md
m-vo: I factored things out into a separate CSS class, made it responsive, and also applied it to another place in ad2d24757ece7fe47ed3d779ed42a586e26e6529.

Usage looks like this now, the arrow gets inserted via a pseudo element:

```md
<div class="from-to">
  <div>From part</div>
  <div>To part</div>
</div>

m-vo:
> please add an empty line after the headlines slightly_smiling_face

Done in 16aa44bd8d4afabd508be064131abc613c11a477.

> I am not sure about the best-practice one. I think all these instances should just be a tip notice (and may be start with Best practice:).

I used the notices first, but they are not suited if you need multiple paragraphs:

image
vs.
image


They also take too much focus, IMHO (large bar, indentation, …). I would even consider toning down the green background some more for the best practice examples.

fritzmg:
> I used the notices first, but they are not suited if you need multiple paragraphs:

Ah yes, that is true.

fritzmg:
Thank you @m-vo :)

PR #1132, opened 2023-01-20
Prioritize new approach to config over legacy approach

bezin:
Hi there, 

I stumbled upon this while reading something up in the docs: The legacy location for config files under `/app/config/` comes before the new approach with everything in `/config`. Wouldn't it make sense to switch this as proposed in this PR? 

It is easy to miss the hint, when you skim through the copy. 

(Would update English accordingly.)

### DIFF
--- a/docs/manual/system/settings.de.md
+++ b/docs/manual/system/settings.de.md
@@ -266,7 +266,7 @@ Datenbankpasswörter, die nur aus Ziffern bestehen oder gewisse Sonderzeichen en
 
 ## config.yml
 
-Die normale Bundle Config gehört in die `config.yml` und befindet sich im Ordner `app/config/`. 
+Die normale Bundle Config gehört in die `config.yml` und befindet sich im Ordner `config/`. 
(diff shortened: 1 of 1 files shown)


fiedsch:
I like the idea, to first describe what is used in the „current version“. 

For „current version“ however, we have to decide what that means (today we have two LTS versions and then there also is Contao 5). I would understand 4.13 LTS as the current version that gets documented as standard (mentioned first).

I‘m curious on other people’s thoughts.

fritzmg (approved):

bezin:
> For „current version“ however, we have to decide what that means (today we have two LTS versions and then there also is Contao 5). I would understand 4.13 LTS as the current version that gets documented as standard (mentioned first).

Yes, good point. I think, it should be the most recent LTS, too.

netzarbeiter:
Thank you, Benedict. It would be nice if you could add the english version.

fritzmg (approved):

netzarbeiter:
I have added https://github.com/contao/docs/commit/750016113f1cd8e6ba36203baa5a8921c50d8535

netzarbeiter (approved):

PR #1136, opened 2023-01-27
a tiny addition

Ernestopheles:
... to make things clear

PR #1138, opened 2023-01-29
clone into active directory

Ernestopheles:
My change will let clone no more into contao/docs but into contao. Then the make commands will work as described from the contao directory.


fritzmg:
Duplicate of #1136

Ernestopheles:
Not a duplicate as different languages

fritzmg:
Could you then please only use one Pull Request? Your changes already explain how to do that ;)

PR #1142, opened 2023-02-01
Do not call news 'messages'

fritzmg:
Currently the English translation of the news extension articles refer to the news posts as "messages". They should be called the same as in the Contao back end, i.e. "news".


netzarbeiter (approved):
Thank you, Firtz

PR #1143, opened 2023-02-06
Use new devilbox domain suffix

Earlopain:
https://github.com/cytopia/devilbox/releases/tag/v3.0.0-beta-0.3

The default suffix for the devilbox has changed from `.loc` to `.dvl.to`, which eliminates the net for manual `etc/hosts` entries. Keep in mind that this change is still in beta, but the instructions don't make mention of downloading a release so I just cloned the repo.


fkaminski:
This ist now mentioned within PR #1129

PR #1146, opened 2023-02-17
Document new operation mechanics

fritzmg:
This adds some basic documentation for the new operation mechanics in Contao 5.

_Note:_ I deliberately did not document (yet) how you could implement a toggle in Contao 4.9 or even 4.13, since it still requires some internal knowledge and I would still recommend to use Haste for toggles in Contao 4 for better DX.


docs/dev/reference/dca/list.md
```diff
@@ -128,3 +128,53 @@ $GLOBALS['TL_DCA']['tl_example']['list']['operations'] = [
...
+    'show',
+    'toggle',
+];
+```
+{{% /notice %}}
+
+
+#### Toggle Operation
+
+{{< version-tag "5.0" >}} You can implement an automatic toggle operation for data containers that contain a boolean
```
Toflar, line 155: That's 4.13 https://github.com/contao/contao/pull/3889

fritzmg: What I documented only works in 5.0. In Contao 4.13 it's still more complicated and I would still recommend to use Haste there (see above).

PR #1149, opened 2023-02-22
Was mir gefehlt hatte...

Ernestopheles:
composer.lock braucht nicht nach meiner Erfahrung. Weitere Ordner und Dateien machen ggf. Sinn.
Hinweis auf .env.

### DIFF
--- a/docs/manual/installation/move-contao.de.md
+++ b/docs/manual/installation/move-contao.de.md
@@ -77,18 +77,28 @@ Gib dein Passwort ein, wenn du danach gefragt wirst.
 {{< /tabs >}}
 
 
+## Einstellungen notieren
+Deine Einstellungen unter `System / Einstellungen` oder im Seitenroot notierst du dir am besten und trägst sie später erneut ein.
+
+{{% notice tip %}}
+Wenn du deine [Einstellungen in einer Umgebungsvariablen ablegst](https://docs.contao.org/manual/de/system/einstellungen/#umgebungsvariablen), entfällt dieser Schritt.
+{{% /notice %}} 
+
 ## Dateien übertragen
 Die folgenden Dateien und Ordner müssen vom Quell- zum Zielserver übertragen werden:
 
 - `files`
 - `templates`
 - `composer.json`
-- `composer.lock`
 
-Falls du noch alte Erweiterungen unter `system/modules` abgelegt hast, alte Konfigurationen unter `system/config`,
-eine `config.yml` im Verzeichnis `config/` (bzw. **vor Contao 4.8** `app/config/`) oder Contao Anpassungen unter
-`contao/` (bzw. **vor Contao 4.8**  `app/Resources/contao/`) angelegt hast, müssen diese auch auf deinen Server
-übertragen werden.
+Die folgenden Dateien und Ordner sollten - sofern genutzt - vom Quell- zum Zielserver übertragen werden:
+
+- `config`  (bzw. **vor Contao 4.8** `app/config/`)
+- `contao`  (bzw. **vor Contao 4.8** `app/Resources/contao/`
+- `src`
+- `.env*`
+
+Falls du noch alte Erweiterungen unter `system/modules` abgelegt hast, alte Konfigurationen unter `system/config` angelegt hast, müssen diese auch auf deinen Server übertragen werden.
 
 Du kannst dazu entweder einen FTP-Client verwenden oder – falls du die Konsole bevorzugst – das Programm `scp`:
 
--- a/docs/manual/installation/move-contao.de.md
+++ b/docs/manual/installation/move-contao.de.md
@@ -93,7 +93,7 @@ Die folgenden Dateien und Ordner sollten - sofern genutzt - vom Quell- zum Ziels
 - `src`
 - `.env*`
 
-Falls du noch alte Erweiterungen unter `system/modules` abgelegt hast oder alte Konfigurationen unter `system/config` angelegt hast, müssen diese auch auf deinen Server übertragen werden.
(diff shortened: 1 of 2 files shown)


docs/manual/installation/move-contao.de.md
```diff
@@ -77,18 +77,28 @@ Gib dein Passwort ein, wenn du danach gefragt wirst.
...
+Wenn du deine [Einstellungen in einer Umgebungsvariablen ablegst](https://docs.contao.org/manual/de/system/einstellungen/#umgebungsvariablen), entfällt dieser Schritt.
+{{% /notice %}} 
+
 ## Dateien übertragen
 Die folgenden Dateien und Ordner müssen vom Quell- zum Zielserver übertragen werden:
 
 - `files`
 - `templates`
 - `composer.json`
-- `composer.lock`
```
fritzmg, line 86: This article is about moving _an existing_ Contao instance _as is_. This means you _must_ copy the `composer.lock`, otherwise you will not get the same dependencies as before when running `composer install`.

docs/manual/installation/move-contao.de.md
```diff
@@ -77,18 +77,28 @@ Gib dein Passwort ein, wenn du danach gefragt wirst.
 {{< /tabs >}}
 
 
+## Einstellungen notieren
+Deine Einstellungen unter `System / Einstellungen` oder im Seitenroot notierst du dir am besten und trägst sie später erneut ein.
```
fritzmg: That is not necessary if you also copy the `system/config/localconfig.php`

docs/manual/installation/move-contao.de.md
```diff
@@ -77,18 +77,28 @@ Gib dein Passwort ein, wenn du danach gefragt wirst.
...
+
+{{% notice tip %}}
+Wenn du deine [Einstellungen in einer Umgebungsvariablen ablegst](https://docs.contao.org/manual/de/system/einstellungen/#umgebungsvariablen), entfällt dieser Schritt.
+{{% /notice %}}
```
fritzmg: Well, not really, you still need to copy the settings over, no? ;)

fritzmg (changes requested):

docs/manual/installation/move-contao.de.md
Ernestopheles: I only copied the composer.json and everything was fine. The dependicies are in the .json, aren't they?

docs/manual/installation/move-contao.de.md
Ernestopheles: I find it misleading if there is said "alte Konfigurationen unter system/config", as I did not use any "old" configuratiuons as fas as I know.

docs/manual/installation/move-contao.de.md
Ernestopheles: Yes, the manual copying drops - just file copying

docs/manual/installation/move-contao.de.md
fritzmg: With the `composer.json` you define the dependencies that you want - in which version range. The `composer.lock` however stores what dependencies are _currently_ installed. If you run `composer install` without a `composer.lock`, then Composer will automatically execute a `composer update` which might result in a different dependency set than you had before. Typically when moving or deploying a Contao instance that is what you do _not_ want. Thus you need to copy the `composer.lock` too.

Ernestopheles: OK, on the other hand: If composer.json is different to the .lock, then with the next update dependencies might be changed anyway.

fritzmg: Not sure what you mean there. If you did not change anything in your `composer.json` before copying the Contao instance to another server, then the `composer.lock` hash will still match the `composer.json`.

docs/manual/installation/move-contao.de.md
fritzmg: Just saying. You can of course improve it ;)

docs/manual/installation/move-contao.de.md
Ernestopheles: I agree to add the composer.lock

docs/manual/installation/move-contao.de.md
Ernestopheles: I hope my changes make it clear now

docs/manual/installation/move-contao.de.md
```diff
@@ -80,15 +80,20 @@ Gib dein Passwort ein, wenn du danach gefragt wirst.
...
+- `system/config/localconfig.php`   (deine Einstellungen)
+
+Die folgenden Dateien und Ordner sollten - sofern genutzt - vom Quell- zum Zielserver übertragen werden:
+
+- `config`  (bzw. **vor Contao 4.8** `app/config/`)         
+- `contao`  (bzw. **vor Contao 4.8** `app/Resources/contao/`
+- `src`
+- `.env*`
+
+Falls du noch alte Erweiterungen unter `system/modules` abgelegt hast oder alte Konfigurationen unter `system/config` angelegt hast, müssen diese auch auf deinen Server übertragen werden.
```
netzarbeiter:
```suggestion
Falls du noch alte Erweiterungen unter `system/modules` oder alte Konfigurationen unter `system/config` angelegt hast, müssen diese auch auf deinen Server übertragen werden.
```

netzarbeiter (changes requested):
Thank you, Ernst

netzarbeiter (approved):

fritzmg (approved):

netzarbeiter (approved):

PR #1154, opened 2023-02-28
Document InsertTagParser service

koertho:
This PR documents the InsertTagParser service. I don't think I've got all necessary information, so additions are welcome :)

### DIFF
--- a/docs/dev/reference/services.md
+++ b/docs/dev/reference/services.md
@@ -524,7 +524,7 @@ class Example
         // response to the client. It will allow ESI tags that can improve caching behaviour.
         $resultChunked = $this->insertTagParser->replaceChunked($buffer);
 
-        // Returns a ChunkedText instaance and should be used when the result is not sent to a client.
+        // Returns a ChunkedText instance and should be used when the result is not sent to a client.
         $resultChunked = $this->insertTagParser->replaceInlineChunked($buffer);
 
         // Example usage for ChunkedText
--- a/docs/dev/reference/services.md
+++ b/docs/dev/reference/services.md
@@ -498,7 +498,9 @@ class Example
 
 This service let you replace inserttags in your string content.
 
-Some methods return a `ChunkedText` instance. The `ChunkedText` container was created to keep the surrounding text containing the insert tags separate from the replacements made by the insert tag parser. It is used for example in the twig escaper to skip encoding on inserttag replacement results.
+Some methods return a `ChunkedText` instance. The `ChunkedText` container was created to keep the surrounding text 
+containing the insert tags separate from the replacements made by the insert tag parser. It is used for example in the 
+twig escaper to skip encoding on inserttag replacement results.
 
 ```php
 use Contao\CoreBundle\InsertTag\ChunkedText;
--- a/docs/dev/reference/services.md
+++ b/docs/dev/reference/services.md
@@ -492,6 +492,51 @@ class Example
 }
 ```
 
+## InsertTagParser
+
+{{< version "4.13" >}}
+
+This service let you replace inserttags in your string content.
+
+
+```php
+use Contao\CoreBundle\InsertTag\ChunkedText;
+use Contao\CoreBundle\InsertTag\InsertTagParser;
+
+class Example
+{
+    private InsertTagParser $insertTagParser;
+
+    public function __invoke(string $buffer): string
+    {
+        // Return a string that may contain esi tags
+        $resultReplace = $this->insertTagParser->replace($buffer);
+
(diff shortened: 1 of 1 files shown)


m-vo:
Some background:

The `ChunkedText` container was created to keep the surrounding text containing the insert tags separate from the replacements made by the insert tag parser. We adjusted the Twig escaper to skip the encoding (i.e. treating it as "raw") for the "raw" chunks. This makes it possible to use different insert tag functions to express which parts should be encoded and which not: https://docs.contao.org/dev/framework/templates/creating-templates/#insert-tags

docs/dev/reference/services.md
```diff
@@ -492,6 +492,51 @@ class Example
...
+use Contao\CoreBundle\InsertTag\ChunkedText;
+use Contao\CoreBundle\InsertTag\InsertTagParser;
+
+class Example
+{
+    private InsertTagParser $insertTagParser;
+
+    public function __invoke(string $buffer): string
+    {
+        // Return a string that may contain esi tags
```
Toflar: I think many people won't understand ESI. I don't think it's super helpful to describe what the methods do. It would be way more valuable to document **when** to use which method. So real world examples.

- `replace()` should be used in HTML context when the replaced result is sent as a response to the client. It will allow for ESI tags that can improve caching behaviour.
- `replaceInline()` should be used when the result is **not** sent to a client. Like e.g. when you use `$email->subject()` or so.

The chunked alternatives have very specific use cases which I think 99% of the developers will never need. But maybe @m-vo or @ausi can come up with a real world example here too?

koertho: The example for chunked texts was taken from the twig ContaoExtension, I just replace twig_escape_filter with htmlspecialchars :)
I will adjust the texts, sound good :+1:

docs/dev/reference/services.md
```diff
@@ -492,6 +492,52 @@ class Example
...
+use Contao\CoreBundle\InsertTag\ChunkedText;
+use Contao\CoreBundle\InsertTag\InsertTagParser;
+
+class Example
+{
+    private InsertTagParser $insertTagParser;
+
+    public function __invoke(string $buffer): string
+    {
+        // Return a string and should be used in HTML context when the replaced result is sent as a response to the client. It will allow for ESI tags that can improve caching behaviour.
```
fritzmg:
```suggestion
        // Return a string and should be used in HTML context when the replaced result is sent as a response to the 
        // client. It will allow for ESI tags that can improve caching behaviour.
```

docs/dev/reference/services.md
```diff
@@ -492,6 +492,52 @@ class Example
...
+        $resultReplace = $this->insertTagParser->replace($buffer);
+
+        // Return a string and should be used when the result is not sent to a client ( e.g. when using `$email->subject()`).
```
fritzmg:
```suggestion
        // Return a string and should be used when the result is not sent to a client 
        // (e.g. when using `$email->subject()`).
```

docs/dev/reference/services.md
```diff
@@ -492,6 +492,52 @@ class Example
...
+        $resultReplaceInline = $this->insertTagParser->replaceInline($buffer);
+
+        // Return ChunkedText instance and should be used in HTML context when the replaced result is sent as a response to the client. It will allow for ESI tags that can improve caching behaviour.
```
fritzmg:
```suggestion
        // Return ChunkedText instance and should be used in HTML context when the replaced result is sent as a 
        // response to the client. It will allow for ESI tags that can improve caching behaviour.
```

fritzmg (changes requested):

fritzmg (approved):

fritzmg:
Thank you @koertho :)

docs/dev/reference/services.md
```diff
@@ -492,6 +492,57 @@ class Example
 }
 ```
 
+## InsertTagParser
+
+{{< version "4.13" >}}
+
+This service let you replace inserttags in your string content.
```
Toflar, line 499:
```suggestion
This service lets you replace insert tags in your string content.
```

docs/dev/reference/services.md
```diff
@@ -492,6 +492,57 @@ class Example
...
+
+Some methods return a `ChunkedText` instance. The `ChunkedText` container was created to keep the surrounding text 
+containing the insert tags separate from the replacements made by the insert tag parser. It is used for example in the 
+twig escaper to skip encoding on inserttag replacement results.
```
Toflar, line 503:
```suggestion
twig escaper to skip encoding on insert tag replacement results.
```

PR #1158, opened 2023-03-06
Minus sign is not allowed

bytehead:
See https://github.com/contao/contao/blob/cd935664582cc3883d8d465fbdf6ecef29b3bce0/core-bundle/src/Resources/contao/library/Contao/Widget.php#L873-L884

### DIFF
--- a/docs/dev/reference/dca/fields.md
+++ b/docs/dev/reference/dca/fields.md
@@ -205,7 +205,7 @@ can be [registered using a hook][3].
 
 | Key         | Description                                                                                                       |
 |:------------|:------------------------------------------------------------------------------------------------------------------|
-| digit       | allows numeric characters only (including full stop [.] and minus [-])                                            |
+| digit       | allows numeric characters only (including full stop [.] and comma [,])                                            |
 | natural     | allows non-negative natural numbers (including 0)                                                                 |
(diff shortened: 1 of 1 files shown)


fritzmg (changes requested):
I cannot confirm that. For example:

```php
// contao/dca/tl_content.php
use Contao\CoreBundle\DataContainer\PaletteManipulator;

$GLOBALS['TL_DCA']['tl_content']['fields']['test'] = [
    'inputType' => 'text',
    'eval' => ['tl_class' => 'w50', 'rgxp' => 'digit', 'maxlength' => 255],
    'sql' => ['type' => 'string', 'length' => 255, 'default' => ''],
];

PaletteManipulator::create()
    ->addField('test', 'text_legend', PaletteManipulator::POSITION_APPEND)
    ->applyToPalette('text', 'tl_content')
;
```

Entering `-10.5` in the `test` field is accepted by the widget (`+10.5` will not be accepted for example).

bytehead:
Of course. I don't know what my brain thought of using this to check of ZIPs with `-`, like `4580-733` 😅

fritzmg:
For that you should use `'rgxp' => 'custom', 'customRgxp' => '[0-9]+-[0-9]+'` (or something like that) 😁

PR #1163, opened 2023-03-23
Fix wrong method usage in Response Context docs

koertho:
This PR change the method name from `getCurrentContext` to `getResponseContext` in JsonLd example code.


fritzmg:
Thank you @koertho

PR #1181, opened 2023-04-14
Add hint for timezone in cronjob for messenger consume

rabauss:
We had some trouble with wrong timezone for async mailing

### DIFF
--- a/docs/manual/system/settings.de.md
+++ b/docs/manual/system/settings.de.md
@@ -978,6 +978,12 @@ bereits erwähnten `--time-limit=1` Option wird der Prozess nach spätestens ein
 [Symfony Dokumentation](https://symfony.com/doc/current/messenger.html#consuming-messages-running-the-worker).
 {{% /notice %}}
 
+{{% notice "note" %}}
+Es kann sein, dass Mails zeitversetzt verarbeitet werden, 
+wenn der Cronjob keine Angabe für die Zeitzone hat und dann den Standard `UTC` verwendet.
+Deshalb sollte die lokale Zeitzone entweder global auf dem Server festgelegt werden oder explizit im Cronjob.
+{{% /notice %}}
+
 
 [SymfonyMailer]: https://symfony.com/doc/4.4/mailer.html#transport-setup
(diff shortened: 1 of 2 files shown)


docs/manual/system/settings.de.md
```diff
@@ -978,6 +978,16 @@ bereits erwähnten `--time-limit=1` Option wird der Prozess nach spätestens ein
...
 
+{{% notice "note" %}}
+Es kann sein, dass Mails zeitversetzt verarbeitet werden, 
+wenn der Cronjob keine Angabe für die Zeitzone hat und dann den Standard `UTC` verwendet.
+Deshalb sollte die lokale Zeitzone entweder global auf dem Server festgelegt werden oder explizit im Cronjob:
+
+```bash
+php -d date.timezone="Europe/Berlin" vendor/bin/contao-console messenger:consume --limit=10 --time-limit=1
+```
+{{% /notice %}}
```
fritzmg, line 985: Not sure that is really relevant? The timezone should of course always be correctly configured in the server environment.

Toflar: Yeah, the timezones should always match. What exactly was the problem?

rabauss: well you won't get any hint and the messages in our case were processed 2 hours later because of `available_at` in the "future" - because on our server the default timezone for php cli was `UTC`

rabauss: we could also remove the example for the cronjob and only add the hint

rabauss: @fritzmg  @Toflar what do you think?

fritzmg: This also applies to `contao:cron` for instance, not just `messenger:consume`. In any case, a hint that the timezone needs to be properly configured would surely not hurt, yes.

rabauss: ok I removed the example cron and just kept the hint

fritzmg (approved):

netzarbeiter (approved):
Thank you, Ralf

PR #1183, opened 2023-04-18
Add list operation example

koertho:
This PR adds an example for list operations to the dca docs.

### DIFF
--- a/docs/dev/reference/dca/callbacks.md
+++ b/docs/dev/reference/dca/callbacks.md
@@ -601,20 +601,23 @@ an additional command check via load_callback).
 
 {{% expand "Example" %}}
 
-This example hides the delete button if user is not an admin.
+This example hide a custom operation button if the user is not allowed to use it.    
+
+Attention: this won't disable the operation itself, it only hides the button!
+To disable the operation, you need to check for the permission additional before it is execution, for example in the operation backend module code.
 
 ```php
 // src/EventListener/DataContainer/ExampleListOperationListener.php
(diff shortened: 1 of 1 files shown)


fritzmg (changes requested):
Unfortunately I don't think this is a good example. This might give the false impression that this prevents non-admins from deleting records. However it only hides the button - and does not prevent the actual action.

koertho:
Ok, I changed the example to a custom button and added additional explanation.
I also enhanced the code with Backend class instead of Controller class and the usage of sprintf.

docs/dev/reference/dca/callbacks.md
```diff
@@ -599,6 +599,64 @@ an additional command check via load_callback).
 **return:** `string` HTML for the button
 {{% /expand %}}
 
+{{% expand "Example" %}}
+
+This example hide a custom operation button if the user is not allowed to use it.    
+
+Attention: this won't disable the operation itself, it only hides the button!
+To disable the operation, you need to check for the permission additional before it is execution, for example in the operation backend module code.
```
fritzmg:
```suggestion
Attention: this won't disable the operation itself, it only hides the button!
To disable the operation, you need to check for the permission additionally 
before its execution, for example in the operation code or a `config.onload` callback.
```

fritzmg (changes requested):

fritzmg (approved):

fritzmg:
Thank you @koertho :)

PR #1184, opened 2023-04-19
Remove width, height, mode from picture insert tag

fritzmg:
The `{{picture::*}}` insert tag does not actually support `width`, `height` and `mode`. It only supports pre-defined image sizes via `size`.

@contao/developers technically we could make it so that the `{{picture::*}}` insert tag also supports these. However it makes no sense to use `{{picture::*}}` for such a use-case anyway.


Toflar (approved):
The parameters for picture indeed do not make any sense.

netzarbeiter (approved):
Thank you, Fritz

PR #1185, opened 2023-04-20
Update config reference dump

ausi:
This is a dump of Contao 4.13 with the new configurations of 5.1 added.

### DIFF
--- a/docs/dev/reference/config.md
+++ b/docs/dev/reference/config.md
@@ -21,46 +21,57 @@ contao:
 
     # The error reporting level set when the framework is initialized. Defaults to E_ALL & ~E_NOTICE & ~E_DEPRECATED & ~E_USER_DEPRECATED.
     error_level:          8183
+    intl:
+
+        # Adds, removes or overwrites the list of ICU locale IDs. Defaults to all locale IDs known to the system.
+        locales:
+
+            # Examples:
+            - +de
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

fritzmg:
Thank you @ausi :)

docs/dev/reference/config.md
```diff
@@ -12,6 +12,8 @@ Contao's configuration options can be reviewed by running the following command:
 $ vendor/bin/contao-console config:dump-reference contao
 ```
 
+The following is a dump of the configuration reference from Contao **5.1**.
```
ausi, line 15: @fritzmg this is not completely accurate as it also includes configuration keys that exist in 4.13 only (like `legacy_routing` for example)
But I thought as a reference having all configurations from 4.13 and 5.1 would be best for the docs.

fritzmg: Oh, so you merged them together? We could also make tabs for Contao 4(.13) and 5(.1). That's easier to maintain.

ausi: > so you merged them together?

Yes.

> We could also make tabs for Contao 4(.13) and 5(.1).

:+1: I will create a new PR for that.

ausi: Done #1191

PR #1188, opened 2023-04-24
Add Latest Version Infos

fkaminski:
An attempt to highlight current version information/changes in the manual. 
A new hugo shortcode "changelog" is used.

The goal would be later to highlight/add some feature highlights here too.

### DIFF
--- a/docs/manual/introduction/_index.de.md
+++ b/docs/manual/introduction/_index.de.md
@@ -12,4 +12,28 @@ Hier lernst du alles, was du zum Arbeiten mit Contao wissen musst. Dazu gehören
 grundsätzlichen Zusammenhänge der Komponenten, die Bedienung des Administrationsbereichs, die verschiedenen 
 Inhaltselemente und Module sowie die Handhabung von Erweiterungen.
 
+
 {{% children %}}
+
+
+## Aktuelle Informationen
+
+
+Ein »[Open Source CMS](/de/einleitung/contao-open-source-cms/)« wie Contao wird ständig weiter entwickelt. Lesenswert sind 
+vor [Migrationen](/de/migration/) die Angaben in den jeweiligen Dateien `changelog` und `upgrade` auf 
+[GitHub](https://github.com/contao/contao). Informationen zum Contao [Update-Zyklus](/de/installation/contao-aktualisieren/#minor-release) sind, 
+neben dem [Release-Plan](https://contao.org/de/release-plan.html), gut dokumentiert.
+ 
+
+### Version 4.13 LTS
+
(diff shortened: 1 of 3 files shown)


fkaminski:
~~Probably we could simplify the regular expression in the "shortcode" - suggestions are welcome.~~

fritzmg:
Hm, so this just shows the changelog from the most recent version?

image

I am not sure how useful this really is. @netzarbeiter wdyt?

fkaminski:
> I am not sure how useful this really is.

Understood and makes sense. 
If such a thing should be desired in principle, 
one could possibly extend the shortcode (show the changes of the last x versions or all within an accordion etc.).

Furthermore, I wanted to add additional targeted FE highlights here 
(ala https://www.youtube.com/watch?v=xH_2XjqM2vc) - roughly. 

How exactly this could look graphically I do not know yet. Something like this:: 
https://portal.productboard.com/7pu88c9kpmqtzt8hwg6arujh/tabs/6-released

fritzmg:
> Furthermore, I wanted to add additional targeted FE highlights here (ala https://www.youtube.com/watch?v=xH_2XjqM2vc) - roughly.
> 
> How exactly this could look graphically I do not know yet. Something like this:: https://portal.productboard.com/7pu88c9kpmqtzt8hwg6arujh/tabs/6-released

What does this have to do with this PR? I am really confused now 😁

fkaminski:
> What does this have to do with this PR?

Not much :-) Just wanted to say that we could later, in addition to the changelog infos, then supplement other version highlights.

fritzmg:
But that already exists on contao.org

fkaminski:
Changed the shortcode. With the parameter "historyCount" (int) you can define the number of versions from the changelog. 
The output then via tabs.

fritzmg (changes requested):
As I said, I am not sure how useful this is. Besides, the information might be out of date, if the docs aren't generated and deployed at the same time a new Contao version is released.

fkaminski:
>  ...if the docs aren't generated and deployed at the same time a new Contao version is released

 True

As you like! I just wanted to update the shortcode in any case. 
If it is not needed - just close it please.

fkaminski:
> But that already exists on contao.org

Only as a supplement. Meant were new highlights/features that were introduced in certain versions (e.g. dark mode/favorites/ajax-Forms in 5.1). 
But as you mentioned. First of all, it has nothing to do with this PR.

netzarbeiter:
This information belongs on the Contao website.
https://contao.org/de/funktionen

fkaminski:
Okidoki!

fkaminski:
With your consent, I would then close this PR, as it is not necessary for the manual.

PR #1189, opened 2023-04-27
Clarify FAQ entry for `public-dir`

fritzmg:
The current FAQ entry for changing the `public-dir` has an error: if you do not change the `DocumentRoot` before executing `composer install` you won't actually be able to execute `composer install` via the Contao Manager (as the Contao Manager would not be reachable).


netzarbeiter (approved):
Thank you, Fritz

PR #1197, opened 2023-05-07
Integrate the meta description

JanoschOltmanns:
We noticed on the Contao camp that the docs are currently delivered without a meta description (see https://github.com/contao/docs/issues/1196). The meta template overwrites the original meta template (see https://github.com/matcornic/hugo-theme-learn/blob/master/layouts/partials/meta.html), so I have added the description again (I don't know if it is possible to implement template inheritance in Hugo ;-)).


netzarbeiter (approved):
Thank you, Janosch

contaoacademy:
As discussed on 09/28/2023, we do not want to integrate this feature this way.

We think it is better to first review where it is even necessary for the page title and the meta title to be different.

We suggest the following approach:

Someone with enough time checks the complete documentation and create a list, where changes to the title are necessary.
If possible, the changes are made directly in the page title
If there is a page title that really needs to be different from the meta title, then create a list from that.

If the list has a significant number of entries, then you can think about implementing it again. Until then, we see no need to implement this.

@fritzmg will close the ticket / PR for now.

PR #1199, opened 2023-05-08
Contao Camp 2023 Session "onboarding" (first draft)

fiedsch:
In the "Onboarding" session we discussed the process of the first Contao installation for new users that have never been
in touch with Contao before.

We assumed they would enter the process via contao.org's download page and follow the steps there until they eventually
(and hopefully) would start reading the documentation via the provided link
[Installationsanleitung](https://docs.contao.org/manual/de/installation/contao-manager/).

Here we spotted some issues, some of which are on the Contao homepage (which @marcuslelle will address) and some in the Contao documentation which will be discussed here:

* Link to the documentation leads to "Über den Contao Manager" (or "About Contao Manager" respectively) whereas a user might expect to find a page that presents something like "Contao installieren".

* We do not want to completely reorganize the documentation's chapter "Installation" as the link from contao.org is not
  the only way to the documentaion and thus should not be the main focus. So we came up with the idea of having a
  "quickstart" page that briefly describes the necessary steps. This page of the documentation could then be the target
  of the link "Installationsanleitung" discussed above.

* Our ideas here:

  - Hide this page in the documentation as it duplicates the content already present in the existing documentation and
    does not provide any additional value for a reader that directly accesses the documentation and browse it via the
    menu.
  - Describe the steps needed in the installation process without going to deep as this might overwhelm new users.
    If the want to learn more they could do so later on with the "regular" installation documentation.
  - Provide the new (potentionally less experienced) users with keywords so the can proprely address issues with
    their web hoster – "I'm having problems installing Contao" will probably not be very helpful there

This is work in progress and is presented here so we have something we could watch and discuss (rather than
"this is exactly the way I want you to merge it, please" ;-)

See also https://github.com/contao-community-alliance/contao-camps/blob/main/2023/camp-2023.md#sonntag-07052023


fiedsch:
"Finalized" by adding the last section.

While working on it I felt, that a shorter version would serve the purpose better. This shorter Version would only contain a list of the required steps, each with links to the existing documentation for detailed information.

What are your opinions on that?

fiedsch:
Added `quickstart-checklist.de.md` which is the above mentioned shorter version.

docs/manual/installation/quickstart-checklist.de.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Den Contao Manager installieren
+
+Der [Contao Manager](../../installation/contao-manager/) besteht aus einer einzelnen Datei, welche über 
+[contao.org](https://contao.org/de/download.html) heruntergeladen werden kann. Nach erfolgreichem Download erhältst 
+du eine Datei `contao-manager.phar`. Übertrage diese Datei in das Verzeichnis `public` auf deinem Webserver.
+
+{{% notice info %}}
+`.phar`-Dateien werden nicht von allen Hosting-Anbietern ausgeführt. Für beste Kompatibilität füge die
+Dateiendung `.php` <b>nach dem Upload</b> hinzu (benenne die Datei <b>auf dem Server</b> in `contao-manager.phar.php` um).
```
fritzmg:
```suggestion
Dateiendung `.php` **nach dem Upload** hinzu (benenne die Datei **auf dem Server** in `contao-manager.phar.php` um).
```

fritzmg (dismissed):
There are now two files with almost the same content. Is that on purpose?

fiedsch:
This is on purpose to present two different approaches, where one would be picked and the other one would be deleted (or both if it is decided not to change the link target on contao.org).

fritzmg:
> This is on purpose to present two different approaches, where one would be picked and the other one would be deleted

Ah I see.

> (or both if it is decided not to change the link target on contao.org).

That does not matter, we can always do a redirect or alias.

fkaminski:
Maybe we could add a link to the following content at the end (?):

https://docs.contao.org/manual/en/guides/add-first-index-page/
respectively
https://docs.contao.org/manual/de/anleitungen/die-erste-startseite/

V15:
> While working on it I felt, that a shorter version would serve the purpose better. This shorter Version would only contain a list of the required steps, each with links to the existing documentation for detailed information.

I like the longer version better.
More details - less open questions.

fiedsch:
Updated docs to include what @fkaminski suggested (Link to "Die erste Startseite")

fritzmg:
@fiedsch did you delete your fork by accident?

fiedsch:
Yes sorry. That probably happened when I tried to get everything back in sync. I still have the local repository. How can I wire things correctly again?

Create an new PR and reference this discussion?

PR #1200, opened 2023-05-08
Fix text to match the screenshot below

fiedsch:
It seems the button's label has been changed and the screenshot is current, but the text is not.


netzarbeiter (approved):
Thank you, Andreas

PR #1202, opened 2023-05-08
Suggestion: Add optional page param for meta title

JanoschOltmanns:
This proposal makes it possible to use an optional parameter `meta_title` in the markdown files, which is placed in the meta-title attribute instead of the `title`. As a fallback, the `title` is retained as before (see https://github.com/contao/docs/issues/1196)

Example:
```
---
title: "Einstellungen"
meta_title: "Contao Einstellungen"
description: ""
url: "system/einstellungen"
aliases:
    - /de/system/einstellungen/
weight: 10
```

Output:
image

Unfortunately, the template of the original theme must be completely overwritten. The adjustment was made according to the example at https://gohugo.io/functions/default/.

### DIFF
--- a/docs/manual/installation/system-requirements.de.md
+++ b/docs/manual/installation/system-requirements.de.md
@@ -1,6 +1,5 @@
 ---
 title: "Systemvoraussetzungen"
-metaTitle: "Systemvoraussetzungen von und mit Contao"
 description: ""
 url: "installation/systemvoraussetzungen"
 aliases:
--- a/docs/manual/installation/system-requirements.de.md
+++ b/docs/manual/installation/system-requirements.de.md
@@ -1,5 +1,6 @@
 ---
 title: "Systemvoraussetzungen"
+metaTitle: "Systemvoraussetzungen von und mit Contao"
 description: ""
 url: "installation/systemvoraussetzungen"
 aliases:
--- a/page/layouts/partials/header.html
+++ b/page/layouts/partials/header.html
@@ -1,0 +1,112 @@
+<!DOCTYPE html>
+<html lang="{{ .Page.Language | default "en" }}" class="js csstransforms3d">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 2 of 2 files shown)


fritzmg (changes requested):
I think it should be `metaTitle`, since we already have `menuTitle`.

Also may be we should look for a solution that does not require overriding the complete `header.html`.

JanoschOltmanns:
> Also may be we should look for a solution that does not require overriding the complete `header.html`.

Just reading the HUGO-docs ;-)

`metaTitle` was my first approach – so I will change it again

JanoschOltmanns:
> Also may be we should look for a solution that does not require overriding the complete header.html.

I will try a PR for the theme. I have not found any other option 😬

contaoacademy:
As discussed on 09/28/2023, we do not want to integrate this feature this way.

We think it is better to first review where it is even necessary for the page title and the meta title to be different.

We suggest the following approach:

Someone with enough time checks the complete documentation and create a list, where changes to the title are necessary.
If possible, the changes are made directly in the page title
If there is a page title that really needs to be different from the meta title, then create a list from that.

If the list has a significant number of entries, then you can think about implementing it again. Until then, we see no need to implement this.

@fritzmg will close the ticket / PR for now.

fritzmg:
To be clear (as there was may be some confusion) - this feature is only necessary if you want the `<title>` be different from the `<h1>` in the main content (marked in red).

image

It is likely that changes to the meta `<title>` do not necessitate having a different `<h1>` as well. If you want the title to be different from the _main menu entry_ (marked in blue) then this is already possible (via `menuTitle`).

If during analysis of the actual content it turns out that we do (at least sometimes) want the `<title>` to be different from the `<h1>` (again - if it only needs to different from the menu title, this is already possible) we can still merge this PR later on. I just want to avoid overriding the whole partial just for this for now (especially since there is no actual change necessary so far).

PR #1203, opened 2023-05-16
Feature/faq schema json ld

Metis77:
This adds a shortcode `faq.html` which replaces the `expand.html`.
It contains markup for faq schema in the JSON-LD format.

Though I am looking for a better approach to not have an independet faq schema on each shortcode, but to build a list of FAQs together in a single schema notation.

### DIFF
--- a/page/layouts/shortcodes/faq.html
+++ b/page/layouts/shortcodes/faq.html
@@ -34,4 +34,4 @@
         }
       }]
     }
-    </script>
+</script>
--- a/page/layouts/shortcodes/faq.html
+++ b/page/layouts/shortcodes/faq.html
@@ -1,0 +1,37 @@
+{{ $_hugo_config := `{ "version": 1 }` }}
+<div class="expand">
+    <div class="expand-label" style="cursor: pointer;" onclick="$h = $(this);$h.next('div').slideToggle(100,function () {$h.children('i').attr('class',function () {return $h.next('div').is(':visible') ? 'fas fa-chevron-down' : 'fas fa-chevron-right';});});">
+        <i style="font-size:x-small;" class="fas fa-chevron-right"></i>
+        <span>
+        {{$expandMessage := T "Expand-title"}}
+    	{{ if .IsNamedParams }}
+    	{{.Get "default" | default $expandMessage}}
+    	{{else}}
+    	{{.Get 0 | default $expandMessage}}
+    	{{end}}
+    	</span>
+    </div>
+    <div class="expand-content" style="display: none;">
+        {{.Inner | safeHTML}}
+    </div>
+</div>
+
+<script type="application/ld+json">
+    {
+      "@context": "https://schema.org",
+      "@type": "FAQPage",
+      "mainEntity": [{
+        "@type": "Question",
+        "name": "{{$expandMessage := T "Expand-title"}}
+            {{ if .IsNamedParams }}
+            {{.Get "default" | default $expandMessage}}
+            {{else}}
+            {{.Get 0 | default $expandMessage}}
+    	{{end}}",
+        "acceptedAnswer": {
+          "@type": "Answer",
+          "text": "{{.Inner | safeHTML}}"
+        }
+      }]
+    }
+    </script>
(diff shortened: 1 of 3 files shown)


page/layouts/shortcodes/faq.html
```diff
@@ -0,0 +1,37 @@
...
+            {{else}}
+            {{.Get 0 | default $expandMessage}}
+    	    {{end}}",
+        "acceptedAnswer": {
+          "@type": "Answer",
+          "text": "{{.Inner | safeHTML}}"
+        }
+      }]
+    }
+    </script>
```
fritzmg:
```suggestion
</script>

```

fritzmg (changes requested):

page/layouts/shortcodes/faq.html
```diff
@@ -0,0 +1,37 @@
...
+    	{{.Get "default" | default $expandMessage}}
+    	{{else}}
+    	{{.Get 0 | default $expandMessage}}
+    	{{end}}
+    	</span>
+    </div>
+    <div class="expand-content" style="display: none;">
+        {{.Inner | safeHTML}}
+    </div>
+</div>
```
fritzmg, line 17: You might not need to duplicate the whole `expand` template here. Have you tried if it works with

```
{{ partial "expand.html" . }}
```

? (or `"layouts/shortcodes/expand.html"`)

Metis77: could not solve this.
seams like `partial` does not work with other shortcodes, but only with partials.

Maybe .parent could help out here ... but could not find a working solution eighter.
Sorry, my fist day with Hugo.
Maybe anybody else does have an idea.

Maybe having a separate html for the faq´s is not the worst ... styling might become different from the expand.html later on.

fritzmg: Yeah it's fine, it was just a thought.

page/layouts/shortcodes/faq.html
```diff
@@ -0,0 +1,37 @@
...
+      "@context": "https://schema.org",
+      "@type": "FAQPage",
+      "mainEntity": [{
+        "@type": "Question",
+        "name": "{{$expandMessage := T "Expand-title"}}
+            {{ if .IsNamedParams }}
+            {{.Get "default" | default $expandMessage}}
+            {{else}}
+            {{.Get 0 | default $expandMessage}}
+    	    {{end}}",
```
fritzmg:
```suggestion
        "name": "{{$expandMessage := T "Expand-title"}}{{ if .IsNamedParams }}{{.Get "default" | default $expandMessage}}{{else}}{{.Get 0 | default $expandMessage}}{{end}}",
```

Otherwise this produces a multi line string. Which technically validates, but I think we should avoid that.

page/layouts/shortcodes/faq.html
```diff
@@ -0,0 +1,37 @@
...
+        "acceptedAnswer": {
+          "@type": "Answer",
+          "text": "{{.Inner | safeHTML}}"
```
fritzmg, line 28: This will produce HTML tags within `text`. Is that valid?

fritzmg (changes requested):

page/layouts/shortcodes/faq.html
Metis77: not sure, if thats valid for "text". 
but that seams to be fine for the hugo default expand.html shortcode.

Also the result / content of the expand / faq should be able to contain html.

fritzmg: > but that seams to be fine for the hugo default expand.html shortcode.

We are discussing the JSON-LD content here, not the regular content.

fritzmg, line 28:
```suggestion
          "text": "{{ trim (.Inner | plainify) "\n" }}"
```

fritzmg (changes requested):

page/layouts/shortcodes/faq.html
Metis77: Oh I see.

Google says on this for the text atttribute: "The answer may contain HTML content such as links and lists ..."
https://developers.google.com/search/docs/appearance/structured-data/faqpage?hl=de#answer

fritzmg: Ah ok cool, then we can leave it as is.

fritzmg (approved):

netzarbeiter (approved):
Thank you, Martin

PR #1182, opened 2023-04-15
Add hugo asset cache busting

fkaminski:
As mentioned in the Slack channel. If a image gets updated the hash should be updated too after hugo build.

PR #1204, opened 2023-05-18
Move images to assets

fritzmg:
Follow up to #1182. This moves all images to assets.


fritzmg:
@netzarbeiter I'll just merge this as is. If we notice any broken images afterwards, we'll just fix them 1 by 1.

PR #1211, opened 2023-06-09
Add DCA Adjustments

fkaminski:
Enable searching for CSS ID/class entries in content elements.


netzarbeiter (approved):
Thank you, Franko

PR #1221, opened 2023-07-26
Update maintenance-template.de.md

greenertux:
Add note about different template path in Contao 5.1.
Apparently the path has changed since 4.9 which is mentioned explicitly in this document.


netzarbeiter (approved):
Thank you, Tristan

PR #1225, opened 2023-08-10
Update callbacks.md

xx81:
If you try to use "onListCustomOperationCallback" as Function name, you will get the error message:

"The contao.callback definition for service XXX is invalid. Either specify a method name or implement the "onButtonCallback" or __invoke method."

Maybe this should be changed.

### DIFF
--- a/docs/dev/reference/dca/callbacks.md
+++ b/docs/dev/reference/dca/callbacks.md
@@ -626,7 +626,7 @@ class ExampleListOperationListener
     )
     {}
 
-    public function onButtonCallback(
+    public function __invoke(
         array $row,
         ?string $href,
         string $label,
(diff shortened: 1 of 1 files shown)


docs/dev/reference/dca/callbacks.md
```diff
@@ -626,7 +626,7 @@ class ExampleListOperationListener
     )
     {}
 
-    public function onListCustomOperationCallback(
+    public function onButtonCallback(
```
fritzmg:
```suggestion
    public function __invoke(
```

Let's use `__invoke` instead so that we don't have to worry about naming the function.

fritzmg (changes requested):

fritzmg (approved):

fritzmg:
Thank you @xx81

PR #1229, opened 2023-08-22
Add information about changing the maintenance template language

koertho:
This PR add a small information how to set the maintenance template language when using the maintenance command.

### DIFF
--- a/docs/manual/guides/maintenance-template.de.md
+++ b/docs/manual/guides/maintenance-template.de.md
@@ -84,6 +84,13 @@ Natürlich lassen sich auch noch weitere Texte wie z. B. die Fußzeile per `XPT.
 WICHTIG: Alle Änderungen wirken sich sowohl beim Wartungstemplate als auch auf alle anderen Errortemplates von Contao aus.
 {{% /notice %}}
 
+Erscheinen die Texte nicht in der korrekten Sprache, muss die gewünschte Sprache als Standard in der Konfiguration festgelegt werden:
+```yaml
+# config/config.yml
+framework:
+  default_locale: de
+```
+
 
 ## Logo anpassen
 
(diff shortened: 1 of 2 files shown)


fritzmg (commented):
I am not sure we should add this. This will change the default locale for the whole application, will it not? Which could have unintended consequences.

koertho:
I think a lot people don't even know about this option and never set it. I also searched a while how to change the language of the maintenance template. Maybe we can adjust the spelling, but this is the points where people will search how to set the language I think.

fritzmg:
> I also searched a while how to change the language of the maintenance template.

There is currently no way to do that. Your config would change the default language for the _whole application_, not just for the maintenance template.

koertho:
I understand the impact. But it is the only way to do that and in the most cases the maintenance template should be outputted in the default language of the website which is often not set as a lot people not aware of this config option. 

What do you think about changing the text to just a small hint and a link to the symfony docs? <https://symfony.com/doc/current/translation.html#configuration> seems to be the correct destination, but I'm not sure if all users know that this options should be added to the config.yaml file.

koertho:
@fritzmg I adjusted the wording and removed the example. Better?

docs/manual/guides/maintenance-template.de.md
```diff
@@ -84,6 +84,7 @@ Natürlich lassen sich auch noch weitere Texte wie z. B. die Fußzeile per `XPT.
 WICHTIG: Alle Änderungen wirken sich sowohl beim Wartungstemplate als auch auf alle anderen Errortemplates von Contao aus.
 {{% /notice %}}
 
+Wird die komplette Contao-Installation über die Kommandozeile in den Wartungsmodus versetzt, wird zur Erzeugung des Wartungstemplates die Standardsprache des des Systems verwendet. Diese kann über die `config.yaml` angepasst werden (`framework.default_locale`), diese Einstellung wirkt sich aber auf das gesamte System aus und kann eventuell zu unerwünschten Nebeneffekten führen.
```
fritzmg:
```suggestion
Wird die komplette Contao-Installation über die Kommandozeile in den Wartungsmodus versetzt, wird zur Erzeugung des Wartungstemplates die Standardsprache des Systems verwendet. Diese kann über die `config.yaml` angepasst werden (`framework.default_locale`), diese Einstellung wirkt sich aber auf das gesamte System aus und kann eventuell zu unerwünschten Nebeneffekten führen.
```

fritzmg (commented):
Yes 👍

fritzmg (approved):

netzarbeiter (approved):
Thank you, Thomas

PR #1232, opened 2023-08-31
reflect removal of {{post::*}}

mwoelke:
Note: It doesn't seem like {{post::*}} was ever given a proper depreciation warning, therefore I didn't add deprecation info.


netzarbeiter (approved):
Thank you, mwoelke

PR #1238, opened 2023-09-25
Update template-data.de.md

Metis77:
To address SEO I would add `variablen` to the text. So `Template-Daten` becomes `Template Variablen` or sometimes an alternative of that.  
This should make the Google search strings `contao template variablen` or `contao tempalte daten` deliver results from the contao docs.


netzarbeiter (approved):
Thank you, Martin

PR #1266, opened 2023-11-27
Update frontend-modules.de.md

Anke:
Added notice that Meta-Fields are no longer available as of C5.


netzarbeiter (approved):
Thank you, Anke

PR #1268, opened 2023-11-28
Use a description of cronjobs that works independent of the versions

Toflar:
The list and periodicity changes between Contao versions so I think for the developers documentation, it's best to give them the tools to find the actual list themselves.

### DIFF
--- a/docs/dev/framework/cron.md
+++ b/docs/dev/framework/cron.md
@@ -3,22 +3,19 @@ title: Cron
 description: "Contao's cron functionality."
 ---
 
+Contao periodically executes some tasks via its own cron functionality. These include mainly cleanup tasks such as
 
-Contao periodically executes some tasks via its own cron functionality. The following
-is a list of tasks executed by Contao's own bundles: 
+* Purge expired comment subscriptions
+* Purge expired registrations
+* Purge expired Opt-In tokens
+* etc.
 
-| Task                                   | Interval |
-|----------------------------------------|----------|
-| Generate calendar RSS feeds            | daily    |
-| Purge expired comment subscriptions    | daily    |
-| Purge temp folder                      | daily    |
-| Purge search cache                     | daily    |
-| Generate XML sitemap                   | daily    |
-| Purge expired registrations            | daily    |
-| Purge expired Opt-In tokens            | daily    |
-| Generate news RSS feeds                | daily    |
-| Purge expired newsletter subscriptions | daily    |
+Because all cronjobs are registered as services and tagged using the `contao.cronjob` tag, you can find all
+cronjobs on your system using the following command:
 
+```bash
+$ vendor/bin/contao-console debug:container --tag contao.cronjob
+```
 
 ## Configuring the Cron Job
 
--- a/docs/dev/framework/cron.md
+++ b/docs/dev/framework/cron.md
@@ -17,6 +17,13 @@ cronjobs on your system using the following command:
 $ vendor/bin/contao-console debug:container --tag contao.cronjob
 ```
 
+{{% notice "note" %}}
+The aformentioned command can also be used in Contao **4.13**. However, this will not find cronjobs that are registered
+via the legacy `config.php` (see below). Unfortunately there is no convenient way in Contao 4 to display registered
+legacy cronjobs. If you want to look these up you could either search for any `$GLOBALS['TL_CRON']` definitions in your 
+Contao instance via your IDE, or use Xdebug for example in order to inspect the `$GLOBALS['TL_CRON']` array.
+{{% /notice %}}
+
 ## Configuring the Cron Job
 
 By default the cron tasks are executed after a response is sent back to the visitor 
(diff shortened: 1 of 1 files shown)


docs/dev/framework/cron.md
```diff
@@ -3,22 +3,19 @@ title: Cron
...
-| Purge expired registrations            | daily    |
-| Purge expired Opt-In tokens            | daily    |
-| Generate news RSS feeds                | daily    |
-| Purge expired newsletter subscriptions | daily    |
+* Purge expired comment subscriptions
+* Purge expired registrations
+* Purge expired Opt-In tokens
+* etc.
 
+Because all cronjobs are registered as services and tagged using the `contao.cronjob` tag, you can find all
```
fritzmg:
```suggestion
Starting with Contao **5** all cronjobs are registered as services and tagged using the `contao.cronjob` tag. Thus you can find all
```

fritzmg (changes requested):

docs/dev/framework/cron.md
Toflar: I wouldn't document this, the legacy crons are also all found there :) You'll need to debug which ones exactly but you can. Right now, there's no info how to do this in 4.13.

fritzmg: Hm not sure what you mean? In Contao 5 you can find all cronjobs this way. There are no legacy Cronjobs in Contao 5.

Toflar: Yeah but why would you add this info? It's the same in 4.13.

Toflar: They are all tagged as `contao.cronjob`. There is just one `LegacyCron` that delegates to the `$GLOBALS` thing. But it's still better to know that than knowing nothing :D

fritzmg: In 4.13 legacy cronjobs still exist, thus you cannot find all cronjobs this way.

Toflar: Yes, you can. You `dump()` in `LegacyCron`.

fritzmg: I meant you cannot find all cronjobs via `debug:container --tag contao.cronjob`.

Toflar: I keep telling you "yes you can".

fritzmg: In Contao 4.13 `debug:container --tag contao.cronjob` will not tell you out of the box which cronjobs are registered, as there can still be legacy cron jobs (and even within Contao there are still some that are registered this way). Someone reading the documentation would assume otherwise, hence my suggestion to change the wording to make sure that everyone understands that this command only reliably tells you the registered cronjobs starting with Contao 5.

Toflar: I disagree. Again, right now, **there is no mention about legacy crons**. Nothing. You have to know about `$GLOBALS['TL_CRON']`. It's not mentioned on this page of the docs at all. So the current status of the docs is: You know **zero** about legacy crons.

With this PR, when running `debug:container --tag contao.cronjob` you will see `LegacyCron` and you will understand that there is a cron and you can see that this one calls `$GLOBALS['TL_CRON']`. 

So in any case this PR is an improvement over the current docs - for all Contao versions.

fritzmg: > I disagree. Again, right now, there is no mention about legacy crons. Nothing. You have to know about $GLOBALS['TL_CRON'].

Legacy cron jobs are documented here: https://docs.contao.org/dev/framework/cron/#registering-cron-jobs

> With this PR, when running debug:container --tag contao.cronjob you will see LegacyCron and you will understand that there is a cron and you can see that this one calls $GLOBALS['TL_CRON'].

I don't think people will draw this inference at all. If anything, we need to document how to also view the legacy cron jobs.

Toflar: I give up then, feel free to do whatever with this PR, sorry but I don't have the energy for this now.

fritzmg: I will add a note on a way to view the cron jobs in Contao 4.13 on top of this change (e.g. adding `dump($GLOBALS['TL_CRON'])` to the `LegacyCron` service like you suggested).

fritzmg:
@Toflar adding `dump($GLOBALS['TL_CRON'])` to the `LegacyCron` won't really do anything when executing `debug:container --tag contao.cronjob`. You could create

```php
// contao/config/config.php
dump($GLOBALS['TL_CRON']);
```

and then execute `cache:warmup`. However, I am not sure we should officially document this - since stuff like that might also show up in `prod`. Overall there isn't really a convenient way to display all registered legacy cronjobs. If we want one, we would need to add one in Contao 4.13.

Toflar:
You can use xdebug, then you see the crons.

fritzmg:
I have added a note in 1ee29b3 - @Toflar wdyt?

Toflar:
Yeah I'm happy with that too.

fritzmg (approved):

fritzmg:
Thank you @Toflar :)

PR #1274, opened 2023-12-09
remove phpMyAdmin notice

Morgy93:
Since https://github.com/ddev/ddev/releases/tag/v1.22.0:

> PhpMyAdmin has been removed from DDEV core, so ddev launch -p no longer works, etc. If you’d like to use PHPMyAdmin, ddev get ddev/ddev-phpmyadmin and use ddev phpmyadmin.

So, the notice about disabling it is not necessary.


Toflar (approved):

Toflar:
Thanks @Morgy93 ❤️

PR #1286, opened 2024-01-06
Change phpmyadmin text paragraph to new ddev version 1.22.0 ddev.de.md

akroii:
Some small adjustments in databse-administration section, which are supposed because of the ddev version change.


netzarbeiter (approved):
Thank you, Andreas

PR #1303, opened 2024-01-29
Document Content Security Policy features

### LINKED ISSUE #1294: CSP framework
Contao 5.3 ships with a superbe CSP framework, which we should document in both, the user as well as the developer manual.

- What is CSP?
- How to enable it
- How to add the nonce to your own scripts in templates
- How to use hashes in templates
- How to have Contao apply hashes to inline styles of WYSIWYG data

fritzmg:


docs/dev/framework/csp.md
```diff
@@ -0,0 +1,229 @@
...
+            /** @var CspHandler $csp */
+            $cspHandler = $responseContext->get(CspHandler::class);
+
+            // Retrieve nonces, add sources, …
+        }
+    }
+}
+```
+
+The following describes the most important methods of the `CspHandler`. In many cases you will want to make these
```
Toflar, line 48: You might want to do that in the code for app stuff. Maybe we should write something along the lines of "in case you are writing an extension you will most likely want to make these adjustments directly from within templates in order to give your users maximum flexibility..."?

docs/dev/framework/csp.md
```diff
@@ -0,0 +1,229 @@
...
+{{% /tab %}}
+{{% tab name="PHP Template" %}}
+```php
+<div style="<?= $this->cspInlineStyle('display:none') ?>">
+```
+{{% /tab %}}
+{{< /tabs >}}
+
+{{% notice "note" %}}
+In order for inline styles to work in browsers that support CSP Level 3 you will also need to add `'unsafe-hashes'` to
```
Toflar, line 150: The hint that this is taken care of for you if you  use `csp_hash()` in Twig or `cspInlineStyle()` in PHP templates is missing. Maybe the hint should go to the tabs themselves? You only need to take care of it in the PHP controller for example.

Toflar (changes requested):
A few nitpicks but otherwise ❤️

docs/dev/framework/csp.md
fritzmg: Not sure, does it really matter? It does not really matter if it's the App or an extension, imho. In both cases you will most of the time only need it in your templates, but also in both cases there might be use cases where you want to directly add sources in your controller for example.

Toflar: Okay, but the info on "why" you would want to do that in the template and not in the controller is missing ;)

fritzmg: Not sure what the answer to that is. Can you make a suggestion on how to word this differently then?

docs/dev/framework/csp.md
```diff
@@ -0,0 +1,229 @@
...
+            $cspHandler = $responseContext->get(CspHandler::class);
+
+            // Retrieve nonces, add sources, …
+        }
+    }
+}
+```
+
+The following describes the most important methods of the `CspHandler`. In many cases you will want to make these
+adjustments directly from your templates though, rather than from within your controller for example. For this reason
```
Toflar:
```suggestion
adjustments directly from your templates though, rather than from within your controller for example. The reason for it being that if you do it from within templates, you can provide a lot more flexibility to your users. If e.g. a part of the template is not output, it also does not need any CSP information for it. Consequently
```

docs/dev/framework/csp.md
Toflar: Done.

docs/dev/framework/csp.md
fritzmg: Updated in fb4d9fc5f2c5050b2413c62b00b3af4cc5eb08bb

Toflar (approved):
❤️

docs/dev/framework/csp.md
fritzmg: I have shortened the wording a bit, wdyt?

Toflar: Perfect, this can be merged :)

PR #1307, opened 2024-01-30
Revised site structure

netzarbeiter:
Page structure revised and moved up one level as a separate navigation point.


fritzmg (approved):
Fixed some errors with some page aliases, otherwise 👍

PR #1350, opened 2024-02-24
Update newsListFetchItems.md

Tastaturberuf:
- more clarity about the return values
- update references

### DIFF
--- a/docs/dev/reference/hooks/newsListCountItems.md
+++ b/docs/dev/reference/hooks/newsListCountItems.md
@@ -33,7 +33,7 @@ be shown in the current request.
 
 Return `false` if this hook should not be considered. Return an integer otherwise. Return `0` if no news entries are found.
 
-If the return value is other than `false`, no further hooks of type `newsListCountItems` will be executed!
+If the return value is anything other than `false`, no further hooks of type `newsListCountItems` will be executed!
 
 ## Example
 
--- a/docs/dev/reference/hooks/newsListCountItems.md
+++ b/docs/dev/reference/hooks/newsListCountItems.md
@@ -33,7 +33,7 @@ be shown in the current request.
 
 Return `false` if this hook should not be considered. Return an integer otherwise.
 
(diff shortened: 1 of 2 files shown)


docs/dev/reference/hooks/newsListFetchItems.md
```diff
@@ -69,4 +71,4 @@ class NewsListFetchItemsListener
 
 ## References
 
-* [\Contao\ModuleNewsList#L200-L215](https://github.com/contao/contao/blob/4.7.6/news-bundle/src/Resources/contao/modules/ModuleNewsList.php#L200-L215)
+* [\Contao\ModuleNewsList#L205-L220](https://github.com/contao/contao/blob/5.3/news-bundle/contao/modules/ModuleNewsList.php#L205-L220)
```
fritzmg:
```suggestion
* [\Contao\ModuleNewsList#L205-L220](https://github.com/contao/contao/blob/5.3.0/news-bundle/contao/modules/ModuleNewsList.php#L205-L220)
```

Otherwise this link will point to the wrong location or result in 404, if either code is changed or the branch gets deleted.

fritzmg (commented):
The same applies to `newsListCountItems` - do you want to add the phrase there as well?

Tastaturberuf:
yes. i made an update to the PR

docs/dev/reference/hooks/newsListCountItems.md
```diff
@@ -33,6 +33,7 @@ be shown in the current request.
 
 Return `false` if this hook should not be considered. Return an integer otherwise.
 
+After an integer has been returned, no further hooks of type `newsListFetchItems` will be executed!
```
fritzmg:
```suggestion
If the return value is other than `false`, no further hooks of type `newsListCountItems` will be executed!
```

I think we should keep the same phrasing (as the check is `=== false`).

fritzmg (changes requested):

fritzmg (approved):

fritzmg:
Thank you @Tastaturberuf

PR #1362, opened 2024-03-20
Update insert-tags.en.md

heimseiten:


fritzmg (commented):
The `{{link::*}}` Insert Tag is meant to be used with page IDs. Does it really work like that?

heimseiten:
Yes, the code written above is what I get. And:
Bildschirmfoto 2024-03-20 um 11 02 25

heimseiten:
But perhaps two examples would also be good. One with a page ID and one with an external URL.

Unfortunately, a phone link doesn't seem to be working. Like: tel:0123456789

fritzmg (changes requested):
I don't think this should be merged as is then. The `*` is a place holder for a Page ID or an absolute URL.

heimseiten:
I agree. Thank you Fritz!

PR #1363, opened 2024-03-22
Update system requirements for Contao 5 under PHP 8.3+

fritzmg:
See https://github.com/contao/contao/issues/6983#issuecomment-2014041710

I decided to simplify the table as you would need to employ the system requirements of Contao 4.13 anyway, if you want to update from 4.4+ to 4.13 for example.


netzarbeiter (approved):
Thank you, Fritz

fritzmg:
@leofeyer @ausi should we merge this? Or wait until we resolved this in Contao itself.

ausi:
> should we merge this? Or wait until we resolved this in Contao itself.

I think we should merge it. We only need to fix it in Contao itself if this causes issues for a larger number of users, right?

leofeyer:
Correct. 👍

PR #1373, opened 2024-04-14
Document `basicEntities` eval option

### LINKED ISSUE #1370: Contao Docs Update: Evaluation basicEntities for Contao 5
As discussed in Slack (https://contao.slack.com/archives/CK4J0KNDB/p1712743045932659), there is probably a new DCA attribute `['eval']['basicEntities']` as of Contao 5.

Please refer to the Contao documentation under:
https://docs.contao.org/dev/reference/dca/fields/#evaluation

fritzmg:
See https://github.com/contao/docs/issues/1370

### DIFF
--- a/docs/dev/reference/dca/fields.md
+++ b/docs/dev/reference/dca/fields.md
@@ -189,7 +189,7 @@ Each field can be validated against a regular expression.
 | uploadFolder       | Path (`string`)                  | The target path for file uploads of the `upload` widget.                                                                                                              |
 | useRawRequestData  | true/false (`bool`)           | If true the raw request data from the Symfony request is used. **Warning:** input filtering is bypassed! Make sure the data is never output anywhere in the back end unescaped which it would if you added the field to a regular back end list view for example. |
 | versionize         | true/false (`bool`)           | If false skip this field in the versioning. Default `true`. |
-| basicEntities | true/false (`bool`) | {{< version-tag "5.0" >}} If true conversts basic entities like `&shy;`, `&amp;` etc. back to their Contao representation `[&]`, `[-]` etc. |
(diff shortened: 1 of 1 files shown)


ausi:
Should we add the info, that in the database they are now always stored as entities?

docs/dev/reference/dca/fields.md
```diff
@@ -188,7 +188,8 @@ Each field can be validated against a regular expression.
 | unique             | true/false (`bool`)           | If true the field value cannot be saved if it exists already.                                                                                                            |
 | uploadFolder       | Path (`string`)                  | The target path for file uploads of the `upload` widget.                                                                                                              |
 | useRawRequestData  | true/false (`bool`)           | If true the raw request data from the Symfony request is used. **Warning:** input filtering is bypassed! Make sure the data is never output anywhere in the back end unescaped which it would if you added the field to a regular back end list view for example. |
-| versionize         | true/false (`bool`)           | If false skip this field in the versioning. Default `true`.
+| versionize         | true/false (`bool`)           | If false skip this field in the versioning. Default `true`. |
+| basicEntities | true/false (`bool`) | {{< version-tag "5.0" >}} If true conversts basic entities like `&shy;`, `&amp;` etc. back to their Contao representation `[&]`, `[-]` etc. |
```
ausi:
```suggestion
| basicEntities | true/false (`bool`) | {{< version-tag "5.0" >}} If true converts basic entities like `&shy;`, `&amp;` etc. back to their Contao representation `[&]`, `[-]` etc. when editing and vice versa when saving. |
```

fritzmg:
> Should we add the info, that in the database they are now always stored as entities?

I think so, yes.

docs/dev/reference/dca/fields.md
fritzmg: That wouldn't be entirely corect, as the vice versa part always happens, regardless of this setting, doesn't it?

ausi: No, both conversions only happen if this setting is activated.

PR #1376, opened 2024-04-19
Update insert tag description for `<acronym>`

zoglo:
Update description for acronym since it's being replaced with `<abbr>`:

See contao:
https://github.com/contao/contao/blob/12a4b8b347fc6d470a4bc6575a7524b4342158a1/core-bundle/src/Resources/contao/library/Contao/InsertTags.php#L898

See deprecated property:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/acronym


fritzmg (approved):

netzarbeiter (approved):
Thank you, Sebastian

PR #1384, opened 2024-05-01
Creates the possibility to display a banner in the head area

netzarbeiter:
To spread the good news of a camp or conference, I have added a banner that can be deactivated and customised via the `params.yml`.

banner

The idea is not mine and was brought to me at the camp.


zoglo:
Wouldn't it look better in the contao colors?

netzarbeiter:
What do you think @zoglo?

banner

zoglo:
Looks better imho

fritzmg (approved):

PR #1396, opened 2024-06-12
Add missing deprecation notices to the insert-tags page

lukasbableck:
Some insert tags which are deprecated / have been removed in https://github.com/contao/contao/commit/719b6cecd813adeb4bf81b25d2ff000cbc9aa124 are not marked as deprecated/removed in the docs.

Also, in the german version of this page there are two different wordings for the deprecations `(seit **Contao 4.13** veraltet, ab **Contao 5.0** nicht mehr verfügbar)` and `(veraltet seit **Contao 4.12**, nicht mehr verfügbar in **Contao 5.0**)`.
Perhaps these should be changed so that they all have the same wording? Personally, I would prefer the first variant.


fritzmg (approved):

netzarbeiter (approved):
Thank you, Lukas

PR #1404, opened 2024-07-08
Update move-contao.de | Add rebuild cache

contaoacademy:
Add notice about rebuilding cache


fritzmg (approved):
In theory this is technically not necessary, if the source of the copy already used `DATABASE_URL`. But it never hurts of course.

contaoacademy:
Thanks for the hint with the DATABASE_URL. However, this does not apply to parameters.yml. Unfortunately, I've had a few support requests recently where the Prod-DB was used by mistake...

netzarbeiter (approved):
Thank you, Christian

PR #1411, opened 2024-07-22
Twig reference initial

fkaminski:
@fritzmg As discussed, a template for Contao's own Twig functions & filters, for the purpose of further content preparation on your part.

### DIFF
--- a/docs/dev/reference/twig.md
+++ b/docs/dev/reference/twig.md
@@ -1,0 +1,61 @@
+---
+title: "Twig Functions and Filters"
+description: Reference of the Contao Twig Functions and Filters.
+aliases:
+    - /reference/twig/
+---
+
+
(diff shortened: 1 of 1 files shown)


fritzmg:
I decided to go a different route #1569
wdyt?

fiedsch:
I like Fritz‘s structure! It keeps the individual pages shorter, so the reader is not overwhelmed, when they start reading (I have the following scenario in mind: a user asks a question in the forum and you provide them with a link to the „solution“, e.g. the filter they should use. Of course you could also link to a section in the middle of a longer page, but then the user seeking help might get confused when reading and scrolling up or down).

fkaminski:
@fritzmg So we could close this draft?

fritzmg:
> @fritzmg So we could close this draft?

If I can interpret your 👍 as agreement, then yes ;)

PR #1415, opened 2024-08-02
Update `contao-component-dir` path for 5.4.x

zoglo:
### Description

This PR updates the `contao-component-dir` to it's initial state as it has been reverted in https://github.com/contao/contao/pull/7406.

Everything regarding the issue about the component-dir can be read within https://github.com/contao/contao/pull/7400

### Quick fix

Since this is a release candidate, this does not need to be documented.
If you are already affected by this, you can simply use the two following commands after reverting the change in your `composer.json`
```
mv /public/assets /assets
```
then run
```
@php vendor/bin/contao-console contao:symlinks
```

### DIFF
--- a/docs/manual/guides/install-test-versions.de.md
+++ b/docs/manual/guides/install-test-versions.de.md
@@ -104,7 +104,7 @@ zu lassen, _inklusive_ den neuesten Release Candidates (wenn vorhanden):
         }
     },
     "extra": {
-        "contao-component-dir": "public/assets"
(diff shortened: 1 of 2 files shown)


zoglo:
There is only 3 occurences since this line has never changed:
https://github.com/contao/docs/pull/1415/files#diff-a82e73cc96359410cb484aad17ee6ec6e9111e2b4c9359bc6bc5c39d083d3c52R166

fritzmg (approved):

PR #1417, opened 2024-08-05
Fix condition example for module listing

kllmanu:
I've just tried the example in a project to only list members of a group. It didn't work as expected and instead listed all members. According to a [forum post](https://community.contao.org/de/showthread.php?2438-Modul-Auflistung-f%C3%BCr-Mitglieder-verwenden-Einschr%C3%A4nken-der-Gruppe&p=15073&viewfull=1#post15073), it is just missing some quotes. 

Sorry for 2 commits, the online "edit this page" feature doesn't seem to allow me to change several files at once.


fritzmg (approved):

fritzmg:
Thank you @kllmanu

PR #1420, opened 2024-08-11
Fix code example

richardhj:
Variable name was ambigous, and unnecessary


fritzmg:
Which variable name was ambiguous?

richardhj:
Sorry, I thought the PaletteManipulator always returns an instance of itself.

Issue #1423, opened 2024-08-14
#top-bar overlapped by #contao-network when scrolled


lukasbableck:
The addition of the #contao-network bar in #1416 causes the top-bar to be overlapped when scrolled down.
This occurs in every major browser.

Example: 
<img width="1632" alt="image" src="https://github.com/user-attachments/assets/094c2f56-02a9-4336-aa05-c9edd89c42c8">

PR #1416, opened 2024-08-03
Add contao.org Top Navigation

fkaminski:
See: #1240
Adopted from contao.org.

PR #1424, opened 2024-08-14
TopNavigation and sticky top-bar

fkaminski:
Quickfix #1423

### DIFF
--- a/page/static/css/theme-contao.css
+++ b/page/static/css/theme-contao.css
@@ -978,3 +978,7 @@ html[data-theme=dark] #toc-menu,
   z-index: 9999;
 }
 
+/* Correction sticky #top-bar */
+.sticky-wrapper.is-sticky > #top-bar {
+  top: 34px !important;
+}


netzarbeiter (approved):
Thank you, Franko

fritzmg:
That makes the top bar jump though.

lukasbableck:
The top bar also overlaps the search overlay a little bit now 

image

fritzmg:
@fkaminski should we revert https://github.com/contao/docs/pull/1416 and https://github.com/contao/docs/issues/1423 instead for the time being?

fkaminski:
> @fkaminski should we revert #1416 and #1423 instead for the time being?

As you like. 
We could fix the `.DocSearch-Modal` class with a custom `top: 34px;` too? 

It would be better in the long term to remove the “`jquery.sticky.js`” and use a pure custom CCS solution. But the “`jquery.sticky.js`” is part of the **theme** partial (footer.html) and it would probably be more difficult to maintain it.

fritzmg:
> We could fix the `.DocSearch-Modal` class with a custom `top: 34px;` too?

I guess that will be sufficient. But I still don't like the sudden jump of the top bar.

fkaminski:
> ... But I still don't like the sudden jump of the top bar.

Me too. But as mentioned, unfortunately we can currently only overwrite (css !important) the inline style from the theme script without touching the theme assets

fritzmg:
Fixed in 65b47d6

PR #1426, opened 2024-08-14
Layout fixes for `#contao-network` topbar

zoglo:
### Description

This should handle all the issues that were recently introduced in #1416 and everything mentioned in #1423 and #1424

* Wrapping the docs content into a `<div id="main">`-container to make sure that `#contao-network` is the only other sibling 17fd5eb48bcfd56f0a6ee029a022ef3800902421 4004d96b925c4f0af12d59542b80a72915a6fb60

* consider FOUC (Flash of unstyled content in dark mode) and CLS (Cumulative Layout Shift) on first visit 6cc4118df79f24b7e6998df71e49d90fa1a342c9

* handle dynamically added nodes (replacement for #1425) df768eec90fc76cbcd50f947e4f8efb4b690a51a

### DIFF
--- a/page/layouts/partials/custom-footer.html
+++ b/page/layouts/partials/custom-footer.html
@@ -44,3 +44,4 @@
 })();
 </script>
 {{ partial "json-ld" . }}
+</main>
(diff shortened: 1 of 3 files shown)


fritzmg (changes requested):
The nav bar still jumps when scrolling down. Also I don't think it is necessary to introduce another HTML element.

fritzmg:
Alternative fix done in 65b47d6

PR #1427, opened 2024-08-15
Fix breadcrumbs tablet view

fkaminski:
The "Diese Seite bearbeiten" (only DE) causes the following display in tablet view. 

image

If we would shorten the text it fits again.

image

### DIFF
--- a/page/i18n/de.toml
+++ b/page/i18n/de.toml
@@ -1,2 +1,2 @@
 [Edit-this-page]
-other = "Diese Seite bearbeiten"
+other = "Seite bearbeiten"


fritzmg:
Would the same issue occur if you go one level deeper?

fkaminski:
> Would the same issue occur if you go one level deeper?

Yepp (even though the breadcrumbs shortens "..."). 

image

Even then its fits with the fix.

image

fkaminski:
Alternatively we could change the media query ?
**from:**

```
@media only all and (max-width: 47.938em) {
  #breadcrumbs .links, #top-github-link-text {
      display: none;
  }
}

```
**to:** (symbol only) 

```
@media only all and (max-width: 47.938em) {
  #breadcrumbs .links {
      display: none;
  }
}

@media only all and (max-width: 64em) {
  #top-github-link-text {
      display: none;
  }
}
```

fritzmg:
I am fine with changing the wording of the link. However, it does not solve the problem completely. The issue is that the theme uses float left/right for these things - instead of flex or grid for example.

netzarbeiter (approved):
Thank you, Franko

PR #1443, opened 2024-10-11
Update deprecated Security service in parseTemplate.md

CMSworker:
New bundle in Symfony 7:
use Symfony\Bundle\SecurityBundle\Security;
Instead of the previous version:
use Symfony\Component\Security\Core\Security;

Old service was deprecated with Symfony 6.2:
https://github.com/symfony/security-bundle/blob/7.1/CHANGELOG.md#62


fritzmg (changes requested):
We can actually change this from `Security $security` to `AuthorizationCheckerInterface $authorizationChecker`. This is then compatible with all Symfony versions.

fritzmg (approved):

fritzmg:
Thank you @CMSworker for bringing this to my attention :)

fritzmg:
Also fixed the other ocurrences in ba936c4

PR #1452, opened 2024-10-29
Correcting method examples/comments and translations

falkgeist:
The static method example for 'findByPk' was named 'findById' (which is a late statically bound method) and the example comment for that method ('findById') was phrased as if it was the 'findByIdAndAlias' static method. Also corrected the translation for "static binded" to "statically bound".

### DIFF
--- a/docs/dev/framework/models/_index.md
+++ b/docs/dev/framework/models/_index.md
@@ -25,7 +25,7 @@ Each class extending `Model` has some *static* methods you can use:
 ```php
 use Contao\PageModel;
 
-// get a record by its primary key
+// get a record by its ID
 $page = PageModel::findById(2);
 
 // find a record by ID or alias (the parameter you pass can be both)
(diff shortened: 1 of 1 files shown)


docs/dev/framework/models/_index.md
```diff
@@ -25,8 +25,8 @@ Each class extending `Model` has some *static* methods you can use:
 ```php
 use Contao\PageModel;
 
-// get a record by its primary key
-$page = PageModel::findById(2);
+// get a record by its primary key (usually the ID)
+$page = PageModel::findByPk(2);
```
fritzmg: The `Model::findByPk()` method is deprecated and should not be used anymore - thus this needs to be reverted.

fritzmg (changes requested):

docs/dev/framework/models/_index.md
falkgeist: Shouldn't we then also remove the method elsewhere in this document (or at least add some info about the deprecation)?

falkgeist: Also, since when is this deprecated? I thought only the use with null is deprecated?

fritzmg: Yes, all `::findByPK` references need to be replaced by `::findById` - which I have done now in 09c0fb2
I don't think we need to add a deprecation info about it in the docs.

falkgeist: Okay, reverted it.

fritzmg (approved):

fritzmg:
Thank you @falkgeist :)

falkgeist:
You're welcome, maybe more from now on.

Issue #57, opened 2019-10-11
Startseite über Hugo erzeugen


fritzmg:
Zusätzliches Environment für die Startseite. Wir könnten da auch einfach das bestehende index.html Template statisch einbinden, denke ich, mit zwei verschiedenen Sprachversionen.

Toflar:
I think when reworking the index page, I would also only link the two books and make them bigger and have some target group related questions or so.

Something like `What would you lilke to read about?`.

-> `I am a user of Contao and I would like to read about how I can use it.`
-> `I am a software developer and I would like to read about how I can extend and work with Contao.`

fritzmg:
The start page is now a redirect to https://contao.org/en/support

PR #1454, opened 2024-11-01
Rework index page

zoglo:
### Description

- redesign + color-changes of the current index page of the docs
- adjusting design to contao.org
- makes sure we don't need a dark-mode for the index page (same for light and dark-mode)

Current state can be checked here:
https://sebastian.oveleon.com/ (en)
https://sebastian.oveleon.com/de.html (de)

___
🙃


fritzmg:
> adjusting design to contao.org

What do you mean here? The design is just a static copy of contao.org. You don't need to adjust things like the border radius of the boxes.

zoglo:
> > adjusting design to contao.org
> 
> What do you mean here? The design is just a static copy of contao.org. You don't need to adjust things like the border radius of the boxes.

I'd rather say match the stage-img article. Could remove the border-radius.

fritzmg:
The design should be kept the same as on contao.org

Otherwise we could create something new entirely anyway (see also https://github.com/contao/docs/issues/57)

PR #1455, opened 2024-11-03
Document how to set up database socket connections

### LINKED ISSUE #584: Document how to use MySQL / MariaDB UDS instead of TCP in parameters.yml
Today i noticed, you can omit `database_host` and `database_port` inside `parameters.yml` to let Doctrine use `/run/mysqld/mysqld.sock`, or what ever is configured in php.ini's mysqli section, instead.

https://github.com/doctrine/dbal/blob/16878650d79934c31e27a13618ed6f3effee80d8/lib/Doctrine/DBAL/Driver/Mysqli/MysqliConnection.php#L67

I assume it would be great to either add `database_unix_socket` and / or `database_dca` to `parameters.yml` in future Contao Versions, as well as document that this Fallback can be used to gain access to the default or php.ini configured UDS.

This Documentation would have saved me quite some Time today.

fritzmg:
Implements #584 

@GottZ can you confirm this is correct?


GottZ:
Sieht gut aus!

fritzmg:
RTM then @netzarbeiter

netzarbeiter (approved):
Thank you, Fritz

PR #1456, opened 2024-11-03
Document legacy routing parameters

### LINKED ISSUE #632: Document "auto item"?
May be someone could cover (at least briefly) such concept as "auto item" in developer's docs? It feels like it is a common knowledge for core developers, but it doesn't seem to actually be documented anywhere...

fritzmg:
Implements #632

This documents the `auto_item` and key/value fragment parameters.


docs/dev/framework/routing/legacy-parameters.md
```diff
@@ -0,0 +1,123 @@
+---
+title: "Legacy Parameters"
+description: "Legacy routing parameters in Contao"
+alias:
+  - /framework/routing/legacy-parameters/
+---
+
+
```
Toflar, line 8: I'm missing like a general warning that clearly states that if you build things from scratch, you should never use `auto_item` anymore but instead the page type :)

fritzmg: Unfortunately I consider this still somewhat experimental, as we don't have an example in the core and there are still issues like https://github.com/contao/contao/pull/7631

We are slowly progressing towards this being the definite way of doing it, but until then, these legacy parameters will still be relevant, even today.

I don't want to put a warning up before we are sure that everything is as it should.

Toflar (approved):

PR #1457, opened 2024-11-03
Document content routing

### LINKED ISSUE #1298: Content URL generator
We have a new content URL generator service that is able to generate URLs for literally everything.
It is the new service to generate URLs for pages, news articles, calendar entries, faq entries etc.

This probably qualifies for an entire new section in the "Framework" chapter.

https://github.com/contao/contao/pull/6596

fritzmg:
Implements #1298


docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
+---
+title: Content Routing
+description: "URL generation for database records in Contao."
+---
+
+
+{{< version "5.3" >}}
+
+In Symfony you can generate the URL to the routes of controllers via the [`router` service][RouterService]. However,
+in Contao you also want to generate front end URLs for your database records (like a news item for example) - or just
```
Toflar:
```suggestion
in Contao you also want to generate front end URLs for your objects (like a news item for example) - or just
```

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+for a regular page.
+
+Contao **5.3** introduces a system with which you can implement "URL resolvers" for your own database records (models,
+entities) and then use the `ContentUrlGenerator` service to generate URLs to these database records - or to any of the
```
Toflar:
```suggestion
Contao **5.3** introduces a system with which you can implement "URL resolvers" for your own objects (models,
Doctrine entities) and then use the `ContentUrlGenerator` service to generate URLs to these database records - or to any of the
```

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    public function __invoke()
+    {
+        $page = PageModel::findBy(…);
+
+        // Generates an absolute URL for the given page
+        $pageUrl = $this->contentUrlGenerator->generate($page, [], UrlGeneratorInterface::ABSOLUTE_URL);
+
+        $news = NewsModel::findBy(…);
```
Toflar, line 46: This should use the framework, I don't think advocating for static calls here is a good idea. Maybe even better would be to have a fragment example which already gets the model in the `__invoke()`.

Toflar (changes requested):

docs/dev/framework/routing/content-routing.md
fritzmg: > This should use the framework, I don't think advocating for static calls here is a good idea.

If you don't need tests I see no reason to use the adapters of the framework. Besides we use `Model::findBy` numerous times in the documentation already. I think it would be superfluous to change this to adapters everywhere.

> Maybe even better would be to have a fragment example which already gets the model in the __invoke().

Hm, what would such an example look like though? Typically you will need to generate the detail URL of an object in a module that lists several such object.

And I am not even sure it is technically possible to have parameters automatically mapped like that anyway?

Toflar: > Hm, what would such an example look like though?

A list controller? That's the whole point of this example?

fritzmg: How would you get the models for a list controller in the `__invoke()` method?

fritzmg: Regardless, I will add an example with a list controller where you generate the URL in the Twig template - that is definitely still missing.

Toflar: I guess based on the settings of the fragment. I mean, in the news list controller I can configure the settings so that's what you have to pass on to your `NewsModel::findBy($criteria)`.

fritzmg: Ah yes ok. Well I have included a simple example now.

Toflar (approved):

docs/dev/framework/routing/content-routing.md
aschempp: > If you don't need tests I see no reason to use the adapters of the framework

Fully agree with this (and stick to this in all of my extensions as well).

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+The news resolver first checks whether the given `$content` is an instance of `NewsModel`. Depending on the type of news
+either of these 3 things will happen next:
+
+* If the news is simply a reference to an external URL, the news URL resolver returns a `ContentUrlResult` with the 
+stored URL as its `StringUrl` content. This will cause the result to be forwarded to the `StringResolver`, which will
+then process the URL further (i.e. replace insert tags and make sure that the URL reference type matches the requested
+type).
+* If the news is a reference to a Contao page or article, a `ContentUrlResult` with the page or article as its content
```
aschempp:
```suggestion
* If the news redirects to a Contao page or article (and not to a news reader), a `ContentUrlResult` with the page or article as its content
```

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+will be returned. This will cause the result to be forwarded to the `PageResolver` or `ArticleResolver` respectively.
+* Otherwise the news resolver determines the target page of the news item's archive and returns a `ContentUrlResolut`
```
aschempp:
```suggestion
* Otherwise the news resolver determines the target page of the news item's archive and returns a `ContentUrlResult`
```

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+with that page as its content.
+
+The difference in the latter case is the usage of `ContentUrlResult::resolve()` rather than 
+`ContentUrlResult::redirect()`. In this case the URL will be resolved for the given target with the current content as
+the target's content.
+
+This is important for the second part of the interface: `getParametersForContent()`. Here your URL resolver can define
+what parameters should be used during URL generation when generating the URL for your content. For example, if your
```
aschempp:
```suggestion
what parameters should be used during URL generation when generating the URL for your resolved content. For example, if your
```

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+The follwing shows a custom example using a page controller. Let's say you have a custom DCA `tl_foobar` and an 
+accompanying `FoobarModel`:
+
+```php
+// contao/dca/tl_foobar.php
+$GLOBALS['TL_DCA']['tl_foobar'] = […];
+```
+
+```php
+namespace App\Model;
```
aschempp, line 238:
```suggestion
// src/Model/FoobarModel.php
namespace App\Model;
```

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+use Contao\Model;
+
+class FoobarModel extends Model
+{
+    protected static $strTable = 'tl_foobar';
+}
+```
+
+```php
+use App\Model\FoobarModel;
```
aschempp, line 250:
```suggestion
// contao/config/config.php
use App\Model\FoobarModel;
```

docs/dev/framework/routing/content-routing.md
```diff
@@ -0,0 +1,301 @@
...
+
+        if (!$request = $this->requestStack->getCurrentRequest()) {
+            return null;
+        }
+
+        /**
+         * This is a simplification for this example. We simply look for the first "foobar_reader" page type in this
+         * website and assume that this is the correct page for which we want to generate the detail URL of our record.
+         */
+        $foobarPage = $this->pageFinder->findFirstPageOfTypeForRequest($request, 'foobar_reader');
```
aschempp, line 321: I would simplify this e.g. by using a property in `$content` as the ID of the target page, removing the complexity of request and pageFinder.

docs/dev/reference/services.md
```diff
@@ -823,6 +823,39 @@ class ExampleService
 }
 ```
 
+## Content URL Generator
+
+{{< version "5.3" >}}
+
+The `contao.routing.content_url_generator` service lets you generate URLs for Contao models and entities (whenever
```
aschempp:
```suggestion
The `contao.routing.content_url_generator` service lets you generate URLs for content objects, e.g. models and entities (whenever
```

aschempp (changes requested):

docs/dev/framework/routing/content-routing.md
fritzmg: That would make the example more complex actually as you would then have to define the reader page somewhere. Either per record (which would be silly) or in a parent record (which this example does not use, i.e. it does not use mode 4).

This is actually how I would implement it in the new `contao/blog-bundle` - you don't _have to_ define a redirect page in the blog archive anymore. If you don't, the system will just look for the first reader page in the page tree. This simplifies the usage.

fritzmg:
Merging, as https://github.com/contao/docs/pull/1457#discussion_r1828632245 is just a detail that can be changed later.

PR #1473, opened 2024-11-14
Change web folder to public

Gabbarowski:
Since 4.13 public is the default folder and fits better with the Symfony environment. I think it would be good to change this. :)


fritzmg (approved):

fritzmg:
Thank you @Gabbarowski :)

PR #1479, opened 2024-11-28
Update namespaces.md

Tastaturberuf:
Make clear that are files not directories


fritzmg (changes requested):
These aren't filenames - these are FQCNs.

Tastaturberuf:
Then lets change the header `example` with `FQCN`? I was not sure what that exactly means.

fritzmg:
You mean from "Example" to "Example Class Name" or "Example FQCN"?

Tastaturberuf:
Sorry seems i'm muted this issue accidentally a while ago. I think `Example FQCN` is a good micro optimization. When i first discover the page, i don't understand it fully.

PR #1487, opened 2024-12-08
Run `backup:create` without touching crontab

### LINKED ISSUE #1485: Add backup:create cron without touching crontab
Can we add an example of how to create backups on a schedule here <https://docs.contao.org/manual/en/cli/db-backups/#have-backups-created-automatically>

**But without** the need to touch the crontab again, when the contao cron command is already in the crontab?

Unsure how to proceed here - maybe as described here <https://docs.contao.org/dev/framework/cron/#asynchronous-cron-jobs>?

richardhj:
Closes #1485


Toflar (approved):
LGTM, although I think this should be an extension :)

netzarbeiter (approved):
Thank you, Richard

PR #1508, opened 2025-02-06
Update content-routing.md

aschempp:
I still think that `$this->pageFinder->findFirstPageOfTypeForRequest($request, 'foobar_reader');` is a bad example. It also only works if the `foobar_reader` page is on the root level of your page tree, which usually is not the case for readers (and only meant for error pages etc.).


fritzmg (approved):

fritzmg:
Feel free to suggest a change for the example - I just mainly wanted to keep the example simple and not go into detail about how your object is connected to a page in the first place.

PR #1514, opened 2025-02-11
[5.5] Document showing the operations in the overview

zoglo:
### Description

With Contao 5.5 around the corner, if you are adding the operations with keys, the default operations will not show within the overview anymore and will be hidden in the context-menu.

This PR documents how to enable those in the short syntax without using `'primary' => true `

image

image

### DIFF
--- a/docs/dev/reference/dca/list.md
+++ b/docs/dev/reference/dca/list.md
@@ -167,7 +167,8 @@ $GLOBALS['TL_DCA']['tl_example']['list']['operations'] = [
 {{% notice "note" %}}
 {{< version-tag "5.5" >}}
 All operations are now shown within the context-menu. If you want them to appear in the overview, you can enable them by
-prepending your key with `!` as seen in the example.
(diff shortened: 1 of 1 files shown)


docs/dev/reference/dca/list.md
```diff
@@ -164,6 +164,24 @@ $GLOBALS['TL_DCA']['tl_example']['list']['operations'] = [
 ```
 {{% /notice %}}
 
+{{% notice "note" %}}
+{{< version-tag "5.5" >}}
+All operations are now shown within the context-menu. If you want them to appear in the overview, you can enable them by
+prepending your key with `!` as seen in the example.
```
fritzmg:
```suggestion
All operations are now shown within the context-menu. If you want them to appear in the overview, you can enable them by
prepending your key with `!` as seen in the example. Note that the operations `edit`, `children` and `toggle` will always appear in
the overview regardless.
```

Issue #1498, opened 2025-01-17
Document background jobs and back end search requirements/options


Toflar:
I will do this, once the docs are compatible with the latest Hugo version :) 

Also see https://github.com/contao/docs/pull/1497

PR #1515, opened 2025-02-12
Document PHP requirements for the back end search

fritzmg:


netzarbeiter (approved):
Thank you, Fritz

Toflar:
Sorry but this is not correct. `sqlite` is not a requirement.

Toflar (changes requested):
We already have an issue for that: https://github.com/contao/docs/issues/1498

fritzmg:
> `sqlite` is not a requirement.

Ah, you mean because it can be configured to something else entirely, right? Yes, then let's wait for the actual documentation :)

PR #1516, opened 2025-02-13
Update list.md

Tastaturberuf:


fritzmg (approved):

fritzmg:
Thank you @Tastaturberuf

Tastaturberuf:
Maybe its a good idea to make a new column for the default values?

fritzmg:
Not sure - I think for the booleans it will almost always be `false` anyway, won't it? This one is one of the exceptions.

Tastaturberuf:
Yeah, but is that crystal clear for all users? I'am also unsure how to describe the the strings or array defaults. They are not `null` because they dosn't exist until you set them.

fritzmg:
> I'am also unsure how to describe the the strings or array defaults. They are not null because they dosn't exist until you set them.

Yeah, that's why I am also thinking that defaults do not always make sense.

PR #1521, opened 2025-02-19
Add $promise to return in async cron

veronikaplenta:
Learned the hard way that `$promise` is null in the static function if you skip this :)


Toflar (approved):

fritzmg:
Thank you @veronikaplenta :)

PR #1533, opened 2025-02-26
Simplify BackendMenuListener example

ameotoko:
This one is more of a suggestion than a fix, feel free to reject.

`KnpMenu` can automatically generate URI to routes, if you pass the route name in the options array as a second argument to `createItem` or `addChild` methods. This allows to slightly simplify the backend menu listener by removing `router` dependency.


fritzmg (approved):
Sure, why not :)

Issue #1535, opened 2025-02-27
Version shortcode displaying content in code block


lukasbableck:
Since 3df4a3af5bd28a83469187386b5134d1f14a8bd2 the version shortcode outputs the content in a code block. /cc @fritzmg 
Example: 
https://docs.contao.org/dev/reference/events/
<img width="1231" alt="Image" src="https://github.com/user-attachments/assets/8b73aa1a-658f-4c90-9d2e-a22d1270f000" />

This is caused by the indentations [here](https://github.com/contao/docs/blob/main/page/layouts/shortcodes/version.html#L12) and [here](https://github.com/contao/docs/blob/main/page/layouts/shortcodes/version.html#L14). 
This is parsed as an indented code block. (https://spec.commonmark.org/0.30/#indented-code-blocks)

Could be fixed by removing the indentation completely, not really pretty, but I don't think there is another way to prevent that.

fritzmg:
Yes, this was just a quick fix. I will remove the indentation for now - but otherwise I think I want to figure out how to render an info box properly - i.e. the same way the theme itself does it.

fkaminski:
Perhaps because of »anything indented by four or more spaces is treated as an indented code block«:
"https://discourse.gohugo.io/t/shortcodes-and-indentation/47088/3"

lukasbableck:
Apparently this could be fixed by using `{{< >}}` shortcodes instead of `{{% %}}`, because the shortcode doesn't need to be parsed as Markdown anyway. Tried it out and it produced the correct result. I can make a PR if you want.
https://gohugo.io/content-management/shortcodes/#notation

![Image](https://github.com/user-attachments/assets/b406a433-43e8-4f6b-80a5-0b8540db942f)

PR #1536, opened 2025-03-01
Change shortcode version call

fkaminski:
See #1535


fritzmg (approved):
With #1537 this is no longer necessary, but we should change it nevertheless for consistency of course.

PR #1538, opened 2025-03-02
Change Search Highlight

fkaminski:
When searching, results are marked as follows (s. Screenshot).

cto-docs-search-highlight-01
cto-docs-search-highlight-02

An alternative suggestion (color only).

### DIFF
--- a/page/assets/css/theme-contao.css
+++ b/page/assets/css/theme-contao.css
@@ -756,6 +756,6 @@ nav.TableOfContents li a code {
 mark {
   all: unset;
   color: #fff;
-  background-color: var(--INTERNAL-PRIMARY-color);
+  background-color: var(--PRIMARY-color);
 }
 
--- a/page/assets/css/theme-contao.css
+++ b/page/assets/css/theme-contao.css
@@ -751,3 +751,11 @@ h6 code,
 nav.TableOfContents li a code {
     font-size: 92%;
 }
+
+/* Change Search Highlight Color */
+mark {
+  all: unset;
+  color: #fff;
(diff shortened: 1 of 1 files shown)


zoglo (changes requested):
that one will definitely clash with headline colors in dark mode as seen here:
image


I would propose to change the `--INTERNAL-ACCENT-color` to the primary color, additionally adding `color: #fff;`

image

fkaminski:
> that one will definitely clash with headline colors in dark mode 
> 
I am not sure about this:
cto-search-highlight-03

But we could change it to `--INTERNAL-ACCENT-color`. 
As you like?

zoglo:
> But we could change it to `--INTERNAL-ACCENT-color`. As you like?

Oh, you are unsetting all the custom styles. `--INTERNAL-ACCENT-color` might be the right usage but w/e then

page/assets/css/theme-contao.css
```diff
@@ -751,3 +751,11 @@ h6 code,
 nav.TableOfContents li a code {
     font-size: 92%;
 }
+
+/* Change Search Highlight Color */
+mark {
+  all: unset;
+  color: #fff;
+  background-color: var(--INTERNAL-PRIMARY-color);
```
fritzmg:
```suggestion
  background-color: var(--PRIMARY-color);
```

Never use `--INTERNAL`.

fritzmg (changes requested):

fritzmg (approved):

netzarbeiter (approved):
Thank you, Franko

zoglo (approved):

PR #1543, opened 2025-03-03
Use `currentContaoVersion` for `install-contao`

fritzmg:
Follow-up on #1539: this would use our `currentContaoVersion` parameter. @netzarbeiter @aschempp wdyt?


aschempp:
fine with me, I don't know what that parameter will contain. I also thought it would render as HTML with background color.

fritzmg:
It's just a parameter that we can use and change in one place. It currently contains `5.5`.

netzarbeiter (approved):
Thank you, Fritz

Issue #982, opened 2022-02-01
Document 4.13 installation in Symfony app


Toflar:
See https://docs.contao.org/dev/getting-started/initial-setup/symfony-application/

Something for you @bytehead?

bytehead:
Yes, why not. You can assign me this.

Toflar:
I have removed the 4.4 chapter in https://github.com/contao/docs/pull/983 already :)

n0-m4d:
hi. any insights on when this documentation will be finished, @bytehead ? is some work on it already done? that could really help me out...

Jerani:
It would be a big help to me too, @bytehead

sheeep:
Related PR https://github.com/symfony/recipes-contrib/pull/1375

fritzmg:
4.13 is actually document - but @bytehead can you now provide a guide for 5.3+? 😁

bytehead:
@fritzmg I'll do 👍🏼

bytehead:
I think this can be closed.

PR #1552, opened 2025-03-10
Add some setup docs for Contao 5.3

bytehead:
See https://github.com/contao/docs/issues/982#issuecomment-2451855220


docs/dev/getting-started/initial-setup/symfony-application/contao-5.3.md
```diff
@@ -0,0 +1,614 @@
...
+
+```
+$ composer require --with-all-dependencies \
+    contao/conflicts:@dev \
+    contao/core-bundle:"5.3.*" \
+    symfony/monolog-bundle:^3.1 \
+    php-http/guzzle7-adapter \
+    terminal42/service-annotation-bundle:^1.1 \
+    toflar/psr6-symfony-http-cache-store:^4.0 \
+    twig/twig:^3.11
```
fritzmg:
```suggestion
    symfony/monolog-bundle:^3.1 \
    php-http/guzzle7-adapter \
    toflar/psr6-symfony-http-cache-store:^4.0 \
    twig/twig:^3.11
```

These are required by the `contao/core-bundle` anyway, so I don't think you would need to manually require them?

bytehead: I always require transient dependencies directly. But I can remove them again if you think this is wrong ☺️

fritzmg: Oh ok, you mean because you would use Twig anyway and directly configure monolog in the App. Then yes you are right.

However, I would at least remove `terminal42/service-annotation-bundle` as that is deprecated anyway in favor of the PHP attributes :)

fritzmg (approved):

fritzmg:
Thank you @bytehead :)

PR #1555, opened 2025-03-14
Update DDEV Guide

fkaminski:
Some tools (e.g. database GUI ) are no longer part of DDEV and can optionally be installed as add-ons


netzarbeiter (approved):
Thank you, Franko.

As i do not use DDEV, I would be glad if someone else could approve this pull request.

fkaminski:
> As i do not use DDEV, I would be glad if someone else could approve this pull request.

For example: https://ddev.readthedocs.io/en/stable/users/usage/database-management/#database-guis
My tests have worked with it.

PR #1559, opened 2025-03-20
Correct code example

cliffparnitzky:


zoglo (changes requested):
The requested changes are going back from a constructor property promotion which is available since PHP 8 (and with the readonly since php 8.1) to an old approach...

Is there any reason you would want to update the docs to be PHP 7.4 compatible?

fritzmg:
There is no need for this change as far as I can see. All code examples in the docs use current modern standards (and Contao's code style).

PR #1560, opened 2025-03-22
Update services.md

Tastaturberuf:
Remove tabs for contao 4 / 5


fritzmg:
Why do you want to remove the tabs?

Tastaturberuf:
Because the code is 99% the same except the use statement. I think it's easier to read instead of hiding it behind a tab.

fritzmg:
I disagree - with the tabs it makes it clear for which version you can use what and it makes the code copy & pasteable.

Besides, depending on the use case it's probably better to inject one of the other two interfaces instead anyway.

PR #1569, opened 2025-04-17
Add a Twig reference

fritzmg:
This adds a Twig reference. Supersedes #1411 

I decided to go a different route and make every Twig filter and function its own page, just like the [Twig documentation](https://twig.symfony.com/doc/3.x/).

This PR initializes the structure and documents the `deserialize` filter and the `content_element` function, serving as a template - they are structured just like in the Twig documentation. All other functions and filters will be yet undocumented, but at least listed.

_Note:_ the user manual part of this PR only adds some tags.

### DIFF
--- a/docs/dev/framework/templates/_index.md
+++ b/docs/dev/framework/templates/_index.md
@@ -2,9 +2,8 @@
 title: "Templates"
 description: Contao's template engine.
 aliases:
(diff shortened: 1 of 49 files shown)


contaoacademy:
🤩

PR #1571, opened 2025-04-19
Document the `escape` Twig filter

fritzmg:
@ausi @m-vo I would need some help here. I want to document the exact differences of the escape strategy that Contao introduces - and also document examples regarding `ChunkedText` from insert tags.


ausi:
The escape differences are described here: https://docs.contao.org/dev/framework/templates/architecture/#double-encoding-prevention

A ChunkedText object contains strings that need to be escaped and other strings that are kept `raw`. So `{{ text|insert_tag_raw|escape('html') }}` only escapes the parts that do not come from the replaced insert tags. However this is not directly related to the Contao escaper, this should work in all Twig templates (not only `@Contao` and `@Contao_*`). See https://github.com/contao/contao/blob/4f485fe5410d7ec1e7c9efc9ee5972844d094aec/core-bundle/src/Twig/Extension/ContaoExtension.php#L282-L292

fritzmg:
> However this is not directly related to the Contao escaper, this should work in all Twig templates (not only `@Contao` and `@Contao_*`). See https://github.com/contao/contao/blob/4f485fe5410d7ec1e7c9efc9ee5972844d094aec/core-bundle/src/Twig/Extension/ContaoExtension.php#L282-L292

But it is still done by overriding Symfony's `escaper` filter, is it not?

fritzmg:
I have merged this as is for now, feel free to make any edits if something is missing or wrong :)

PR #1577, opened 2025-04-26
Add FAQ entry for adjusting HTML attributes of headlines

fritzmg:
Since this was already asked a few times in the community.


zoglo (approved):
Looks good to me, might be a bit confusing with the trailing and leading whitespace trimming but it's an example that could be copied.

I guess using parent() is due to modifications from other bundles since there is no content inside it in the original template?

fritzmg:
> might be a bit confusing with the trailing and leading whitespace trimming

Confusing in what way? Without it your (dev) output will be

```html
<h2         class="foobar"
>
```

> I guess using parent() is due to modifications from other bundles since there is no content inside it in the original template?

There is content inside the block: https://github.com/contao/contao/blob/f3c7f25292fe3ccd1680c6c27b5b87cf1c3ea240/core-bundle/contao/templates/twig/component/_headline.html.twig#L22

Without `{{ parent() }}` this example would not work.

zoglo (approved):

netzarbeiter (approved):
Thanky you, Fritz

PR #1615, opened 2025-06-25
Update insert_tag.md

FlowinBeatz:
Added a third example for variable usage

### DIFF
--- a/docs/dev/reference/twig/functions/insert_tag.md
+++ b/docs/dev/reference/twig/functions/insert_tag.md
@@ -14,6 +14,9 @@ template. This can be useful for executing specific functionalities that are onl
 
 {# Insert an article #}
 <p>{{ insert_tag('insert_article::123')|raw }}</p>
+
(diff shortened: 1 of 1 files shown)


fritzmg (commented):
Not sure this is the right place - since this is a [basic Twig feature](https://twig.symfony.com/doc/3.x/templates.html#other-operators), unrelated to the `insert_tag` function. And I think it might be confusing as it is not clear what the `email` variable is or where it came from.

FlowinBeatz:
As Contao always had a heart for casual developers I see no problem in over explaining things. This could at least have saved one support post on Slack :-)

fritzmg:
But it's not well explained and not clear why such an example exists for this function. It would be more suitable to mention concatenation may be [here](https://docs.contao.org/dev/framework/templates/getting-started/#learning-the-syntax) instead.

fritzmg:
I have mentioned concatenation now in 91aa1c4a.

Issue #1474, opened 2024-11-14
Document the removal of config options from 4.13 -> 5.x 


zoglo:
_TL;DR - `url_suffix`, `legacy_routing` or similar when updating_

People are repeatedly asking the same question on the community forums (sometimes by the same person), that they can not update because of a long error message including:

```
  Unrecognized option "url_suffix" under "contao". Available options are "bac  
  kend", "backup", "console_path", "crawl", "cron", "csp", "csrf_cookie_prefi  
  x", "csrf_token_name", "editable_files", "error_level", "image", "insert_ta  
  gs", "intl", "localconfig", "locales", "mailer", "messenger", "pretty_error  
  _screens", "preview_script", "sanitizer", "search", "security", "upload_pat  
  h".  
  ```

Either I am missing the part where configuration options have been removed (unless they are on specific pages as a hint) but it may help if that could be documented on this page
https://docs.contao.org/manual/de/migration/#contao-4-13-auf-5-x

Opening the issue since I am unsure if that is the right place or not

fritzmg:
Mentioning that you might have to remove some config options when updating from Contao 4 to 5 should definitely be done in the migration article.

The config article needs to have two (or more) tabs for the config options.

zoglo:
> The config article needs to have two (or more) tabs for the config options.

Would be a separate issue / not part of this one but can you tell me what you are referring to on https://docs.contao.org/dev/reference/config/?

fritzmg:
The tabs here:

* https://docs.contao.org/manual/en/system/settings/#config-yaml
* https://docs.contao.org/dev/reference/config/

They already exist (but might need to be updated).

zoglo:
Don't mind assigning me to this issue then :)

zoglo:
https://github.com/contao/docs/issues/1474#issuecomment-2475781300
> The tabs here:
> 
> * https://docs.contao.org/manual/en/system/settings/#config-yaml
> * https://docs.contao.org/dev/reference/config/
> 
> They already exist (but might need to be updated).

See #1486 for the `but might need to be updated`

zoglo:
See #1619

PR #1619, opened 2025-07-06
Document the removal of configuration options

zoglo:
### Description

* #1474 

Checked the `config.yaml` of 4.13 and 5.x and only saw `contao.legacy_routing` and `contao.encryption_key`.
I did however add `contao.url_suffix` and `contao.prepend_locale` as these do not exist anymore and are the usual culprits as well.


fritzmg:
My plan is still to create a version switch for 4.x / 5.x - so I'd rather not document it this way.

fritzmg (approved):
But then again, this is for the _Migration_ article anyway, so it makes sense.

netzarbeiter (approved):
Thank you, Sebastian

PR #1629, opened 2025-07-21
Use a site param for the release candidate version

zoglo:
### Description

* same as https://github.com/contao/docs/pull/1496 but now easier to change :)
* introduces a site param `currentReleaseCandidate`

_Yes, if you copy the composer.json via the copy button, it will be the 5.6 version ;)_ 

#### Before

image

#### After

image

PR #1633, opened 2025-08-07
Update modify-container-at-compile-time.md

Tastaturberuf:


zoglo (approved):
Since the symfony docs always work with a version tag, wouldn't it be better using a site param like here to update those urls?

https://github.com/contao/docs/pull/1629

Or would that not work with the future version toggle 🤔

fritzmg:
> Or would that not work with the future version toggle 🤔

It would work - the version toggle is nothing special. It will just link to differently built sites.

Tastaturberuf:
I thought the same thing, a variable one would be very practical.

PR #1644, opened 2025-09-06
Add missing parameter to `csp_hash` Twig function

lukasbableck:
The `csp_hash` function, as it uses the `CspRuntime`s `addHash` method, expects at least two arguments, the directive and the source. 
In the documentation currently only the second argument is passed, which wont work.


fritzmg (approved):

fritzmg:
Thank you @lukasbableck :)

PR #1651, opened 2025-09-12
Fix ue umlaut

de-es:
Replace multibyte char


Toflar:
Without checking the source, what did you do here? Replace the decomposed with the precomposed form of the `ü` maybe? If so, why would we care?

ausi:
We should normalize all the text to `NFC` form, not only the `ü` characters.

Toflar:
Indeed, they are not NFC, that's weird 😕 Is it only the `ü`'s in question here?

fritzmg:
> Is it only the `ü`'s in question here?

I did a quick check and it seems to be only the `ü`, but I'll do some more digging afterwards.

de-es:
Ja, es war nur beim `ü` ein Problem. Die anderen Umlaute waren i.O.

PR #1657, opened 2025-09-24
Introduce version switch

fritzmg:
This introduces a version switch between 5.x and 4.x. Initially the `main` branch will deploy the `5.x` version - once this PR is merged I will create a separate `4.x` branch for the `4.x` version.

Afterwards we can remove 4.x-only stuff from the `main` branch and 5.x-only stuff from the `4.x` branch.

The new URL scheme will be

* https://docs.contao.org/5.x/dev/
* https://docs.contao.org/5.x/manual/en/
* https://docs.contao.org/5.x/manual/de/
&nbsp;
* https://docs.contao.org/4.x/dev/
* https://docs.contao.org/4.x/manual/en/
* https://docs.contao.org/4.x/manual/de/

I will merge this after the conference - so please voice any thoughts or objections until then 🙂


Toflar:
I agree that we need a solution for the version cluttering but we decided against multiple branches because it makes contributions for people not familiar with git a lot more complicated.
What are the plans exactly here? One version per major version of Contao?

fritzmg:
> I agree that we need a solution for the version cluttering but we decided against multiple branches

Yes, but over time it got too complicated to keep track of all the differences between 4.x and 5.x - and we decided against removing all the 4.x stuff outright, as there are of course still _many_ Contao installations running on 4.x.

> because it makes contributions for people not familiar with git a lot more complicated.

Nothing changes regarding contributions.

> What are the plans exactly here? One version per major version of Contao?

Correct.

PR #1670, opened 2025-09-29
Make LTS version a site parameter

Toflar:
Noticed some hardcoded 4.x references and thought that it would be good making those dynamic to config. Also updated the config to represent the current state of things.


page/config/_default/params.yml
```diff
@@ -13,8 +13,9 @@ search:
 disableLanguageSwitch: true
 documentation_label: Documentation
 disableInlineCopyToClipBoard: true
-currentContaoVersion: '5.5'
-currentReleaseCandidate: '5.6'
+currentContaoLtsVersion: '5.3'
+currentContaoVersion: '5.6'
+currentReleaseCandidate: '5.7'
```
fritzmg: This won't work - it will then display `5.7@RC` in the docs, which you will not be able to install.

fritzmg (changes requested):
Ah, I thought I already changed this at those spots. Though imho we don't really need to overcomplicate it with so many different versions - installing should just reference the newest stable version imho.

I'll adjust the PR accordingly.

page/config/_default/params.yml
Toflar: Kk, feel free to update that too then

PR #1679, opened 2025-10-02
Update update-contao.en.md to Contao 5.6

FlowinBeatz:
Changed mentions from Contao 5.5 to Contao 5.6


zoglo (changes requested):
We could and should use the SiteParam 

https://learn.netlify.app/en/shortcodes/siteparam/

https://github.com/contao/docs/pull/1670

fritzmg:
We do not have a site param for the current bugfix version though. I think that should be removed, since it will always be out of date.

And may be we need a param for the LTS version after all, like originally suggested by @Toflar

zoglo:
> And may be we need a param for the LTS version after all

👍 

We would technically need two😁. `5.3` and `5.7` will be LTS simultaneously, at least for one year (Same as 4.13 being a former LTS but still being supported :) )

fritzmg:
No, once 5.7 is out, this will be the LTS version. There is no point creating a _new_ contao installation with 5.3 when 5.7 is released.

docs/manual/installation/update-contao.en.md
```diff
@@ -28,13 +28,13 @@ may not work anymore. The current major release of Contao is **version 5** when
 
 A minor release is a kind of milestone on the development path where new features have been added. Minor changes to 
 existing pages might therefore be necessary. When writing these lines, the current minor version of Contao 
-is **version 5.5**.
+is **version 5.6**.
```
fritzmg:
```suggestion
is **version {{% siteparam "currentContaoVersion" %}}**.
```

docs/manual/installation/update-contao.en.md
```diff
@@ -115,7 +115,7 @@ If you are updating for a [minor release](#minor-release), you need to specify t
 {
     …
     "require": {
-        "contao/manager-bundle": "5.5.*",
+        "contao/manager-bundle": "5.6.*",
```
fritzmg:
```suggestion
        "contao/manager-bundle": "{{% siteparam "currentContaoVersion" %}}.*",
```

fritzmg (changes requested):

fritzmg (approved):

netzarbeiter (approved):

PR #1680, opened 2025-10-03
Update first-bundle.md

Tastaturberuf:
don't do that in C5 bundles


fritzmg:
> don't do that in C5 bundles

Depends on your bundle though and is not related to the Contao version.

Tastaturberuf:
we should only teach the best practice in my opinion

fritzmg (approved):

PR #1682, opened 2025-10-07
Update callbacks.md

Tastaturberuf:
We need this information anymore? Everybody should start with at least with a 5.3 version.

### DIFF
--- a/docs/dev/reference/dca/callbacks.md
+++ b/docs/dev/reference/dca/callbacks.md
@@ -355,8 +355,6 @@ window.
 
 ### `config.onpalette`
 
-{{< version "5.3" >}}
-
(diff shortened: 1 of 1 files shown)


fritzmg:
I deliberately left it in, yes. And there would be more than just these 2 occurrences.

fritzmg:
> Everybody should start with at least with a 5.3 version.

The tags are there so that you know which features came in which Contao version. The documentation does not make any assumption about "what you start with". You might have inherited an old Contao 4.4 or 4.9 instance for example and then it might be useful to know which settings or features are not yet available there if you need to change something quickly, before updating.

Tastaturberuf:
Ok, make sense for me.

PR #1691, opened 2025-10-29
Add example 2

zonky2:
I often see that the entire template is copied in Tiny... :(

### DIFF
--- a/docs/manual/layout/templates/php/template-inheritance.de.md
+++ b/docs/manual/layout/templates/php/template-inheritance.de.md
@@ -20,7 +20,8 @@ angegeben und ihre Inhalte überschrieben werden.
 
 Mittels `$this->parent()` lässt sich der originale Inhalt des Blocks ausgeben.
 
-#### Beispiel 1
+#### Beispiele
+
 Das Template `fe_page.html5` ist in mehrere Blöcke aufgeteilt (u.&nbsp;a. `head`, `meta`, `body`, `footer`).
 Wir möchten lediglich eine weitere Meta-Angabe hinzufügen – dazu schreiben wir unser Template wie folgt:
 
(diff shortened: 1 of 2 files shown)


docs/manual/layout/templates/php/template-inheritance.de.md
```diff
@@ -32,3 +32,25 @@ Wir möchten lediglich eine weitere Meta-Angabe hinzufügen – dazu schreiben w
   <meta name="author" content="Max Muster">
 <?php $this->endblock(); ?>
 ```
+
+#### Beispiel 2
+Möchte man den [TinyMCE-Editor](../../anleitungen/tinymce-konfiguration/) grundlegend oder eigene Ableitungen
+anpassen, so sollten nur die entsprechenden Blöcke überarbeitet werden werden. Um zu ermitteln, welche
+Blöcke zur Verfügung stehen, sieht man sich am besten das Originaltemplate `be_tinyMCE.html5` an.
+Folgend ein Beispiel für eine Anpassung als `be_tinyMCE_Redakteure.html5`:
```
fritzmg: This requires additional knowledge about DCA adjustments. If you just want to generally customize the TinyMCE, you should just adjust the `be_tinyMCE.html5` template. For the purpose of this example I would not create a variant that you'd have to specifically enable then in the respective DCAs.

fritzmg (changes requested):

docs/manual/layout/templates/php/template-inheritance.de.md
zonky2: That's why I linked to the TinyMCE page...

fritzmg: Yes, but there is no point for this simple example to make it more complicated than it needs to be in my opinion.

zonky2: @netzarbeiter wdyt?

netzarbeiter: I would suggest replace that example and adapting the text as follows.

Möchtest du am TinyMCE Anpassungen vornehmen, sollten nur die entsprechende Blöcke überschrieben werden. Um zu ermitteln, welche Blöcke dir zur Verfügung stehen, siehst du dir am besten das Originaltemplate `be_tinyMCE.html5` an.

Mit folgender Anpassung wird die Formatierung vom Text vor dem Einfügen entfernt.

```php
<?php $this->extend('be_tinyMCE'); ?>

<?php $this->block('custom'); ?>
  paste_as_text: true,
<?php $this->endblock(); ?>
 ```

zonky2: fixed

fritzmg (approved):

fritzmg (approved):
@netzarbeiter this is ready to be merged from my point of view 👍 🙂

netzarbeiter (approved):

PR #1696, opened 2025-11-14
Add FAQ entry for Turbo + Browser Extension incompatibility

fritzmg:
This adds an FAQ entry for a fairly common issue that can occur in the back end since Contao 5.4.


zoglo (approved):
> that modify the DOM in an invalid way

Just a nit here but we could also say it more descriptive:
"that inject code into the DOM in an invalid way"

fritzmg:
I would keep "modify" as that would include anything (changing, removing and adding).

fritzmg:
@netzarbeiter ping for review

netzarbeiter (approved):

PR #1701, opened 2025-12-02
Change template type in Content Elements & Moduls

echtermax:
Use required `Contao\CoreBundle\Twig\FragmentTemplate` instead of the deprecated and no longer permitted `Contao\Template` as an argument for the `getResponse()`-function in order to extend the `AbstractFrontendModuleController`


zoglo (approved):
The HTML5 templates are still permitted / allowed but not recommended :)

fritzmg (approved):

fritzmg:
Thank you @echtermax - I also changed all other occurrences 👍

PR #1711, opened 2026-01-17
Remove note saying `getFrontendModule` hook is not executed for modules included via the Module content element

lukasbableck:
Since Contao 5.0 the hook is also executed when a module is included via the Module content element, as it uses `Controller::getFrontendModule` since https://github.com/contao/contao/pull/5001.
So the note is not correct and IMO can be removed completely.


fritzmg (approved):

fritzmg:
Thank you @lukasbableck :)

PR #1712, opened 2026-01-26
Fix template path in back-end route documentation

aschempp:


zoglo (approved):

fritzmg:
This change isn't really necessary, is it? Within your `App` any template within `templates/` can be rendered without a namespace declaration.

aschempp:
That didn't work for me in Contao 5.7 🤔 
Maybe @m-vo can clarify?

fritzmg:
I just dropped the given example in the docs verbatim into Contao 5.7 and it worked.

m-vo:
The Symfony filesystem loader also has a concept of namespaces denoted by `@`, but by default, it internally stores anything in the root template directory under the `__main__` namespace and normalizes names without a `@` to `@__main__/<name>`. So this should work in general. 

But if the rendered template is in fact meant to be a Contao template, it must also include the `@Contao`, otherwise none of the features of the Contao Filesystem Loader will be available. Another argument for renaming `<project-dir>/templates` to `<project-dir>/user-templates` in Contao 6 btw. to end the confusion about Symfony/Contao templates.

aschempp:
The problem most likely was that I place application templates in `/contao/templates` and not the root folder. It only works if you place it exactly in the root templates directory. Which also wouldn't be the case when developing a custom back end controller in a bundle, right?

fritzmg:
Since this is a normal Symfony Controller the expectation is that you place it in the `templates/` folder, regardless of whether it is for the App or a Bundle.

aschempp:
so you don't think it is better to have a documentation that _just works_ instead of _it works if you follow XY without having XY explained_? 🤷 feel free to close then.

fritzmg:
I don't understand, it does work the way it is documented. If we added this it wouldn't work for bundles anymore, so why would we add this?

aschempp:
what wouldn't work for bundles anymore?

fritzmg:
Putting the template into `templates/` as instructed, but then using `@Contao/` for rendering. Only the App shares the `templates/` folder with regular Symfony templates.

PR #1714, opened 2026-02-05
Update jobs framework docs for 5.7

Toflar:
This adds the rest of all the stuff that was planned for 5.7 of our new jobs framework.

I did not bother to write about the differences between 5.6 and 5.7 because 5.6 is dead in a few days anyway and I think it would only confuse people.

### DIFF
--- a/docs/dev/framework/jobs.md
+++ b/docs/dev/framework/jobs.md
@@ -3,7 +3,7 @@ title: Jobs
 description: "Contao's jobs framework."
 ---
 
-{{< version "5.6" >}}
+{{< version "5.7" >}}
 
 {{% notice "warning" %}}
 The entire jobs framework is currently considered *experimental* and therefore not covered by Contao's BC promise. Classes marked with `@experimental` should be considered internal for now.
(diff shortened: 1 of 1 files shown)


docs/dev/framework/jobs.md
```diff
@@ -112,12 +112,85 @@ $job = $job->withMetadata(['api_key' => 'foobar', 'iteration_offset' => 11]);
...
+        if (!$job || $job->isCompleted()) {
+            return;
+        }
+        
+        // Mark a job pending as soon as you start processing it:
+        $job = $job->markPending();
+        $this->jobs->persist($job);
+        
+        // In this example, the total is unknown, but we want to show progress to the user.
+        foreach ($items as $i => $item) {
```
fritzmg: Should we write something like

```suggestion
        foreach ($message->getItems() as $i => $item) {
```

here (`$items` is undefined otherwise), so that it's clear the the actual data to be processed by the `MessageHandler` is still stored in the `Message` itself.

Toflar: That would be certainly wrong, you don't store items on a message, that would make it likely way too big.

fritzmg: The actual data might be stored elsewhere for sure - the items could just be database IDs for instance of course. Nevertheless, the main point is that `$items` is undefined in this example.

Toflar: Better?

fritzmg (approved):
The new example creates a new nuance 😁 - ideally you should first fetch the items and then close the database connection, before processing the items, if the work your `MessageHandler` has to do takes very long for one or each item. This is because the idle timeout configured in the database server might close the connection automatically in between, which will then lead to an error if there are other database operations in between (the idle timeout can be comparatively short).

But it's fine for now to illustrate handling the `Job` instance.

PR #1719, opened 2026-02-11
Fix syntax errors in `add_schema_org` and `attrs` Twig function examples

lukasbableck:
Using `=>` to assign values and `[...]` is not valid syntax for Twig mappings.


fritzmg (approved):

fritzmg:
Thank you @lukasbableck :)

PR #1725, opened 2026-02-25
Add explanation for web URL cron scope and limitations

contaoacademy:
The `### Web URL` section currently only shows how to trigger cron jobs via `_contao/cron`, but does not explain what this actually means conceptually or what its limitations are.

This PR adds a short explanation clarifying that:
- The web URL trigger is conceptually equivalent to a regular website visit
- Cron jobs restricted to the CLI scope (e.g. `contao.cron.supervise_workers`) are not executed via this route
- Each cron job receives a `$scope` parameter and can decide how to behave (see Scope section)
- The 30-second PHP execution time limit applies in the web context
- Setting up a real CLI cron job is recommended for reliable execution of all jobs

### DIFF
--- a/docs/dev/framework/cron.md
+++ b/docs/dev/framework/cron.md
@@ -100,6 +100,26 @@ you could use the following instructions for example:
 * * * * * wget -q -O /dev/null https://example.org/_contao/cron
 ```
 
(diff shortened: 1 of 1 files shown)


fritzmg (approved):

fritzmg:
Thank you @contaoacademy :)

PR #1726, opened 2026-02-25
Clarify which cron jobs run on website visit

contaoacademy:
The article currently explains that cron jobs run on every website visit and recommends setting up a real cron job for performance reasons. However, it does not mention that the web-triggered cron also has functional limitations.

This PR adds a notice block (DE+EN) clarifying that cron jobs triggered by a website visit or the `_contao/cron` route only execute jobs designed for that trigger. The back end search index, for example, is only built via a real CLI cron job. This gives users an additional functional reason to set up a real cron job, beyond the performance argument already covered.


fritzmg (approved):

netzarbeiter (approved):
Thank you, Christian

PR #1729, opened 2026-02-26
Update Twig templates overview for Contao 5.7 (DE+EN)

contaoacademy:
As of Contao 5.7, every `.html5` template has a Twig equivalent and Twig is now the default standard. The section \"Twig templates in Contao core\" was updated to reflect this:

- Updated the opening paragraph to state that as of 5.7, every `.html5` template has a Twig equivalent and Twig is now the new standard
- Added explanation of the template resolution logic: Twig is used by default; a `.html5` file in the `templates` folder takes precedence
- Added a concrete example (`news_full.html.twig` / `news_full.html5`)
- Made the transition period paragraph more specific, referencing legacy `ce_*.html5` elements
- Strengthened the warning to explicitly recommend against HTML5 templates and legacy `ce_*` content elements
- Removed the outdated paragraph stating that not every element has a Twig template yet
- Minor fix: `.html.twig` file extension (dot was missing)

### DIFF
--- a/docs/manual/layout/templates/twig/_index.de.md
+++ b/docs/manual/layout/templates/twig/_index.de.md
@@ -9,14 +9,14 @@ tags: [Twig]
 ---
 
 {{% notice note %}}
-Der gesamte Abschnitt behandelt die Verwendung von Twig-Templates in Contao ab Version 5.0.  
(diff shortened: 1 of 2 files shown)


zoglo (approved):

netzarbeiter (approved):
Thank you, Christian

PR #1731, opened 2026-02-26
Document the `Password` widget

fritzmg:


docs/dev/reference/widgets/password.md
```diff
@@ -1,6 +1,35 @@
...
+
+## Example
+
+```php
+// …
+'myPassword' => [
+    'inputType' => 'password',
+    'eval' => [
+        'mandatory' => true,
+        'preserveTags' => true,
```
ausi:
```suggestion
        'useRawRequestData' => true,
```

Should we use `useRawRequestData` here?

fritzmg: Hm, that's how we use it in `tl_user` at least.

ausi (approved):

PR #1739, opened 2026-03-01
Remove outdated database configuration

fritzmg:
This removes the documentation on how to switch from `utf8mb4` to `utf8`. This is legacy and won't even work on modern MySQL servers anymore (current versions do not ship with `utf8` = `utf8mb3` support). Contao 6 will not be compatible with it either so I think we should just remove it entirely in the Contao 5 documentation already.


netzarbeiter (approved):
Thank you, Fritz

PR #1747, opened 2026-03-27
🚬

Tastaturberuf:
👍


fritzmg (changes requested):
I cannot reproduce that. That callback should never be executed in the front end. You would only need that for a `config.onsubmit` callback for `tl_member`: https://github.com/contao/contao/blob/270c8e2147ea15fd617fbc05cf283771f5dfccdc/core-bundle/contao/modules/ModulePersonalData.php#L376-L390

fritzmg:
Your callback is for `tl_member`, not for `tl_news`.

fritzmg:
See the explanation above - the change does not make sense.

PR #1749, opened 2026-03-31
Document notEditable

koertho:
This PR add notEditable to the dca field reference.


zoglo (commented):
I am not aware of a `fields` option for `notEditable` (not in the docs and not in code), maybe you are confusing it with the option for the `config` key?

https://docs.contao.org/5.x/dev/reference/dca/config/#reference

Image

fritzmg:
@koertho For (some) fields there is [`readonly`](https://docs.contao.org/5.x/dev/reference/dca/fields/#:~:text=readonly), is that what you mean?

koertho:
I thougth the one mentioned here: https://github.com/contao/contao/pull/8838
But maybe the config one was meant?

I searched for a possbility to have a non sql field not be written to the json storage

fritzmg:
`notEditable` has nothing to the with the virtual fields. As already mentioned `notEditable` is a DCA `config` option (so that you cannot edit the records of this DCA).

koertho:
Ok, I had an additional look in the code and you're absolutely right 🫥
Forget about it :)

PR #1763, opened 2026-05-14
Clarify `parameters.yaml`

fritzmg:
The documentation still states that the `parameters.yaml` is created automatically - which hasn't been the case for a long time. It's also not the main way to store database or SMTP credentials anymore.

This PR removes the relevant info and describes the `parameters.yaml` in a more general way. We could move it down, but different sections on this page still mention the `parameters.yaml` alternative, thus I left it at its original place for now.


netzarbeiter (approved):
Thank you Fritz

PR #1764, opened 2026-05-21
Add guide for custom Stimulus controllers

falkgeist:


zoglo:
Unsure if I like having this in the documentation since people don't really need to use stimulus but can use mutation observers. ~~Especially not within the dev as it's not part of the Contao API.~~

falkgeist:
I see your point.

The problem with mutation observers is, that any changes to the DOM based on the observers might in turn trigger Stimulus' observer 🤪. But that is a different issue, imho.

The issue described in the first paragraph is very real though, if you ask me:

Custom backend JS with listeners to the `DOMContentLoaded` event is only triggered on 'real' page loads, not Turbo navigation.

Implementing your own controller is not the definitive fix for that (there's your mentioned mutation observers or also Turbo events), but it does mitigate the issue.

docs/dev/guides/adding-stimulus-controllers.md
```diff
@@ -0,0 +1,91 @@
+---
+title: "Adding Stimulus Controllers"
+description: "How to add custom Stimulus controllers for the back end."
+aliases:
+  - /guides/adding-stimulus-controllers/
+---
+
+Since Contao 5 now uses Turbo and Stimulus for the backend, sometimes it might not be enough to register JS assets for the backend as described under [Adding Backend assets][AddingBackendAssets], especially when listening for the `DOMContentLoaded` event.
```
fritzmg, line 8: The use of Stimulus by the Contao back end is irrelevant here - only the use of Turbo matters. But even that is not completely correct. Even the legacy JavaScript of the Contao back end might add, remove or modify DOM nodes, so it is always a good idea to write JavaScript that can handle any kind of DOM manipulation.

docs/dev/guides/adding-stimulus-controllers.md
```diff
@@ -0,0 +1,91 @@
...
+
+In this case it might be necessary to add a custom Stimulus controller.
```
fritzmg, line 10: It's not necessary, but Stimulus makes it way easier to implement what I mentioned above.

docs/dev/guides/adding-stimulus-controllers.md
```diff
@@ -0,0 +1,91 @@
...
+    
+    beforeCache() {
+        this.#cleanup();
+    }
+    
+    #cleanup() {
+        // revert the markup before caching and when disconnecting
+    }
+    
+
```
fritzmg, line 43:
```suggestion
```

docs/dev/guides/adding-stimulus-controllers.md
```diff
@@ -0,0 +1,91 @@
...
+### Adding your Stimulus application
+
+```js
+// assets/backend.js
+import { Application } from "@hotwired/stimulus";
+import MyCustomWidgetController from "./controllers/my-custom-widget-controller";
+
+const application = Application.start();
+application.register("my-vendor-prefix--my-custom-widget", MyCustomWidgetController);
```
fritzmg, line 55:
```suggestion
application.register('my-vendor-prefix--my-custom-widget', MyCustomWidgetController);
```

docs/dev/guides/adding-stimulus-controllers.md
```diff
@@ -0,0 +1,91 @@
...
+// Note that Stimulus' disconnect() function will not fire at this point and
+// thus cannot be used for this task.
+document.documentElement.addEventListener('turbo:before-cache', (e) => {
+    for (const controller of application.controllers) {
+        if ('function' === typeof controller.beforeCache) {
+            controller.beforeCache(e);
+        }
+    }
+});
+```
```
fritzmg, line 68: I wonder if we really need to include this bit here. I think it would be better to keep it short and simple and just refer to the internals [here](https://docs.contao.org/5.x/dev/internals/_stimulus-backend/). Whether you want something to be executed before the page is being put into the cache really depends on the controller. Ideally a simple controller should not need to care.

@m-vo wdyt?

PR #1766, opened 2026-05-27
Fix linked images

de-es:
Fix Hugo warning: `WARNING you must call the ref / relref shortcode with '% %' instead of '< >' to work correctly for the anchor target attribute`


fritzmg (changes requested):
I don't think this will work in the production build - and it wouldn't have any cache busting.

fritzmg:
Generally I consider this a bug or limitation in Hugo. It's an inline link to an image, that's why we have to use `<` `>` instead of `%` - otherwise the text would be rendered as a paragraph.

de-es:
:+1:  Ich hatte auch schon in der Hugo- und Theme-Doku nach einer Lösung gesucht, aber nichts Passendes gefunden. Dann lassen wir es besser erstmal so.

PR #1767, opened 2026-06-01
Properly apply `algolia` dark mode when using the auto-theme with `prefers-color-scheme: dark`

zoglo:
### Description

Has been bugging me since we switched to the relearn theme :D

This applies dark mode style when in auto mode and prefers color-scheme dark.
Custom properties taken from https://cdn.jsdelivr.net/npm/@docsearch/css@3

### Screenshots

**prefers-color-scheme: dark - Auto**
grafik

**prefers-color-scheme: dark - Light**
Bildschirmfoto 2026-06-01 um 23 59 17

### DIFF
--- a/page/layouts/partials/custom-header.html
+++ b/page/layouts/partials/custom-header.html
@@ -1,6 +1,28 @@
 <meta name="docsearch:manualtype" content="{{ .Site.Params.algolia_selector }}{{ if eq .Site.Params.algolia_selector "manual" }}_{{ .Page.Language | default "en" }}{{ end }}">
 <link rel="preconnect" href="https://KDZKB9K935-dsn.algolia.net" crossorigin>
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3">
+<style>
+    @media screen and (prefers-color-scheme: dark) {
+        html[data-r-theme-variant="auto-theme"] {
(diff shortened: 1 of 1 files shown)


fritzmg (commented):
Shouldn't we put this into `theme-contao-dark.css` instead?

zoglo:
> Shouldn't we put this into `theme-contao-dark.css` instead?

I was unsure at first but yes, can do.

zoglo:
@fritzmg sadly doesn't work as the hugo relearn theme compiles it, thus removed it for the auto-theme when used in `dark` or `light`. Even using `theme-contao` won't work as it will compile into something like this:

```css
/* More margin for footer menus */
#R-sidebar .R-shortcutmenu-footercontrols > ul {
    display: flex;
    flex-direction: column;
    gap: 0.125rem;
}
```

grafik

fritzmg:
Well no, the issue is that https://cdn.jsdelivr.net/npm/@docsearch/css@3 uses the `data-theme` attribute for its dark mode styles. Thus we currently have a JavaScript that sets this attribute depending on the selection. But it does not work properly for `auto`.

You could remove that JavaScript and provide the necessary styles in `theme-contao-dark.css` instead. Or you fix the JavaScript so that it also works for `auto`.

fritzmg:
Fixed in e5a941b4 - the JavaScript checked for the wrong value.

PR #1758, opened 2026-05-06
Update the Row Wizard documentation

zoglo:
### Description

- Removes the known limitation for `eval.dcaPicker` -> See https://github.com/contao/contao/pull/9810
    ~~(Might want to merge this once 5.7.5 has been released)~~
- Updates the picture and removes the note about the saveCallback for emptying values
    (Might want to merge this once ~~[#9843](https://github.com/contao/contao/pull/9843) has been merged and~~ Contao 5.7.6 has been released)

PR #1771, opened 2026-06-05
Document Row Wizard cell_class and callbacks

koertho:
This PR enhance the row wizard docs with information about cell_class and field callbacks.

### DIFF
--- a/docs/dev/reference/widgets/row-wizard.md
+++ b/docs/dev/reference/widgets/row-wizard.md
@@ -56,32 +56,8 @@ This table only shows the options relevant to the core functionality of this wid
 
 ## Field callbacks
 
-You can register callbacks for Row Wizard fields using the following schema: `fields.[rowWizardField].fields.[field].[callback_name]`.
-
-Example:
-
-```php
-namespace App\EventListener\DataContainer;
-
-use Contao\CoreBundle\DependencyInjection\Attribute\AsCallback;
-use Contao\DataContainer;
-use HeimrichHannot\SmulRegionalportalBundle\Model\MemberModel;
-
-#[AsCallback(table: 'tl_example', target: 'fields.exampleWizard.fields.awesomePeople.options')]
-class AwesomePeopleOptionsListener
-{
-    public function __invoke(?DataContainer $dc = null): array
-    {
-        $options = [];
-        $members = MemberModel::findBy(['exampleProperty=?'], ['awesome']);
-        foreach ($members ?? [] as $member) {
-            $options[$member->id] = $member->firstname . ' ' . $member->lastname;
-        }
-        return $options;
-    }
-}
-```
-
+You can register callbacks for Row Wizard fields using the following schema: `fields.<WIZARD_FIELD>.fields.[FIELD].[CALLBACK_NAME]`, 
+for example `fields.rowWizard.fields.type.options`
 
 ## Column Definition
 
--- a/docs/dev/reference/widgets/row-wizard.md
+++ b/docs/dev/reference/widgets/row-wizard.md
@@ -48,15 +48,17 @@ This table only shows the options relevant to the core functionality of this wid
 |----------------------------------|--------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
 | `inputType`                      | `rowWizard` (string)     |                                                                                                                                                    |
 | `fields`                         | `array`                  | Associative array of DCA field definitions rendered per row.                                                                                       |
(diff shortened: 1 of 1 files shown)


zoglo:
I agree with `cell_class`, I disagree about the callback. We recently removed it here:
https://github.com/contao/docs/pull/1758

We already say it here, maybe it needs to be more prominent:
grafik

Either way, if you document an options callback, you'd also need to document save callbacks and every other callback possible for widgets, which I don't think is necessary.

koertho:
@zoglo I disagree. My callback documentation is about callback for the inner fields, the part you removed was a callback for the row wizard field itself. Since the registration differs (`fields.exampleWizard.fields.awesomePeople.options` instead of `fields.awesomePeople.options`), we should at least leave a note.

zoglo:
Maybe we keep it more general and explain that you can target the fields inside by using the `fields.yourWizard.` prefix? A full blown example may not be needed, meaning simply the service tag should be fine with a reference to callbacks. /cc @fritzmg

koertho:
I remove the full example and updated the text.

docs/dev/reference/widgets/row-wizard.md
```diff
@@ -44,14 +44,20 @@ Additional information about known limitations can be read in the associated [pu
...
-| `eval.max`      | `int`                | Maximum number of rows.                                                    |
+| Key                              | Value                    | Description                                                                                                                                        |
+|----------------------------------|--------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| `inputType`                      | `rowWizard` (string)     |                                                                                                                                                    |
+| `fields`                         | `array`                  | Associative array of DCA field definitions rendered per row.                                                                                       |
+| `fields.[field].eval.cell_class` | CSS class(es) (`string`) | Same as [`eval.tl_class`](https://docs.contao.org/5.x/dev/reference/dca/palettes/#arranging-fields)but the classes are applied to the widget cell. |
```
fritzmg:
```suggestion
| `fields.<FIELD>.eval.cell_class` | CSS class(es) (`string`) | Same as [`eval.tl_class`]({{% relref "fields#evaluation" %}}) but the classes are applied to the widget cell. |
+
| `fields.<FIELD>.eval.cell_style` | CSS styles (`string`) | Same as [`eval.style`]({{% relref "fields#evaluation" %}}) but the styles are applied to the widget cell. |
```

docs/dev/reference/widgets/row-wizard.md
```diff
@@ -44,14 +44,20 @@ Additional information about known limitations can be read in the associated [pu
...
+| `eval.actions`                   | `array`                  | Allowed values: `copy`, `delete`, `enable`. Default: `['copy', 'delete']`.                                                                         |
+| `eval.sortable`                  | `bool`                   | Enables or disables drag & drop sorting (default: `true`).                                                                                         |
+| `eval.min`                       | `int`                    | Minimum number of rows.                                                                                                                            |
+| `eval.max`                       | `int`                    | Maximum number of rows.                                                                                                                            |
+
+## Field callbacks
+
+You can register callbacks for Row Wizard fields using the following schema: `fields.<WIZARD_FIELD>.fields.[FIELD].[CALLBACK_NAME]`,
```
fritzmg:
```suggestion
You can register [callbacks]({{ relref "callbacks" }}) for Row Wizard fields using the following schema: `fields.<FIELD>.fields.<FIELD>.<CALLBACK>`, 
```

docs/dev/reference/widgets/row-wizard.md
```diff
@@ -44,14 +44,20 @@ Additional information about known limitations can be read in the associated [pu
...
+| `fields.[field].eval.cell_class` | CSS class(es) (`string`) | Same as [`eval.tl_class`](https://docs.contao.org/5.x/dev/reference/dca/palettes/#arranging-fields)but the classes are applied to the widget cell. |
+| `eval.actions`                   | `array`                  | Allowed values: `copy`, `delete`, `enable`. Default: `['copy', 'delete']`.                                                                         |
+| `eval.sortable`                  | `bool`                   | Enables or disables drag & drop sorting (default: `true`).                                                                                         |
+| `eval.min`                       | `int`                    | Minimum number of rows.                                                                                                                            |
+| `eval.max`                       | `int`                    | Maximum number of rows.                                                                                                                            |
+
+## Field callbacks
```
fritzmg:
```suggestion
## Field Callbacks
```

fritzmg (changes requested):

zoglo (approved):

fritzmg:
Thank you @koertho :)

PR #1774, opened 2026-06-12
Document cron return values

koertho:
This PR add return values to the cron example since returning other values throw an exception since some newer versions

### DIFF
--- a/docs/dev/framework/cron.md
+++ b/docs/dev/framework/cron.md
@@ -152,7 +152,7 @@ use Contao\CoreBundle\DependencyInjection\Attribute\AsCronJob;
 #[AsCronJob('hourly')]
 class ExampleCron
 {
-    public function __invoke()
+    public function __invoke(): void
     {
         // Do something …
     }
@@ -177,7 +177,7 @@ use Contao\CoreBundle\ServiceAnnotation\CronJob;
  */
 class ExampleCron
 {
-    public function __invoke()
+    public function __invoke(): void
     {
         // Do something …
     }
@@ -209,7 +209,7 @@ namespace App\Cron;
 
 class ExampleCron
 {
-    public function __invoke()
+    public function __invoke(): void
     {
         // Do something …
     }
@@ -284,7 +284,7 @@ use GuzzleHttp\Promise\PromiseInterface;
 #[AsCronJob('hourly')]
 class HourlyCron
 {
-    public function __invoke(string $scope): PromiseInterface
+    public function __invoke(string $scope): ?PromiseInterface
     {
         // Skip this cron job in the web scope
         if (Cron::SCOPE_WEB === $scope) {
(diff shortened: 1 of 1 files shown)


docs/dev/framework/cron.md
```diff
@@ -284,7 +284,7 @@ use GuzzleHttp\Promise\PromiseInterface;
 #[AsCronJob('hourly')]
 class HourlyCron
 {
-    public function __invoke(string $scope): PromiseInterface
+    public function __invoke(string $scope): ?PromiseInterface
```
fritzmg:
```suggestion
    public function __invoke(string $scope): PromiseInterface
```

The method never returns `null`.

docs/dev/framework/cron.md
```diff
@@ -317,7 +317,7 @@ class HourlyCron
 {
     public function __construct(private ProcessUtil $processUtil) {}
 
-    public function __invoke(string $scope): PromiseInterface
+    public function __invoke(string $scope): ?PromiseInterface
```
fritzmg:
```suggestion
    public function __invoke(string $scope): PromiseInterface
```

The method never returns `null`.

fritzmg (changes requested):

fritzmg:
> since returning other values throw an exception since some newer versions

What exactly do you mean by that? If you mistakenly had a return value other than `GuzzleHttp\Promise\PromiseInterface`, it would simply be ignored.

docs/dev/framework/cron.md
koertho: Null is allowed: https://github.com/contao/contao/blob/d7ca93b28dad1c30b4982cd2b167c9fd8467b4a0/core-bundle/src/Cron/CronJob.php#L50

koertho:
> > since returning other values throw an exception since some newer versions
> 
> What exactly do you mean by that? If you mistakenly had a return value other than `GuzzleHttp\Promise\PromiseInterface`, it would simply be ignored.

Since 5.3.41 the return value is checked and a exception is thrown if not valid: https://github.com/contao/contao/pull/8738

koertho:
(which should have been a breaking change IMHO :grin:   )

fritzmg:
> (which should have been a breaking change IMHO 😁 )

When introduced, CronJobs did not support any return values. Since Contao `5.1.0` support for a `PromiseInterface` return value for asynchronous CronJobs was introduced. Validating these does make sense.

koertho:
It make sense, but broke one of our extensions. But that's not the point of this topic, I already fixed it and wanted to document it :)

docs/dev/framework/cron.md
fritzmg: The method itself never returns `null`.

fritzmg (approved):

fritzmg:
Thank you @koertho

PR #1775, opened 2026-06-19
Fix the field callback parameters

leofeyer:
As discussed in https://github.com/contao/contao/pull/9860, all the field callbacks should provide a `DataContainer` object and never `null`, so this PR adjusts the documentation accordingly.

| Callback | docs allow `\|null` |
| --- | --- |
| `attributes` | yes (now fixed) |
| `options` | yes (now fixed) |
| `input_field` | no |
| `load` | no |
| `save` | no |
| `wizard` | no |
| `xlabel` | no |

### DIFF
--- a/docs/dev/reference/dca/callbacks.md
+++ b/docs/dev/reference/dca/callbacks.md
@@ -750,7 +750,7 @@ Allows you do dynamically adjust the attributes of a field in a DCA before a wid
 
 {{% expand "Parameters" %}}
 * `array` Current attributes
-* `\Contao\DataContainer`/`null` Data Container object
+* `\Contao\DataContainer` Data Container object
 
 **return:** `array` The adjusted attributes array
 {{% /expand %}}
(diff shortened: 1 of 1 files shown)


fritzmg (commented):
Are we really sure it cannot be `null`, ever? I have made a quick test in `5.3` and it seems ok though. (For a `tl_member` `options_callback` you might have to use `DataContainer|Module` though.)

leofeyer:
Yes, that's why we fixed it in https://github.com/contao/contao/pull/9860.

It is also pretty inconsistent at the moment. Two of seven field callbacks randomly permit `|null`, even though it is never actually passed. The others don't. Doesn't make much sense to me.

fritzmg (changes requested):
But it is definitely not guaranteed to not be `null`. See https://github.com/contao/contao/blob/00ac2a1baf2e9db08891db3ef3b07d61e402757f/core-bundle/contao/library/Contao/Widget.php#L1223-L1235

`$objDca` allows `DataContainer|Module|null` and its default is `null`. So if someone executes `Widget::getAttributesFromDca()` in their own code and does not pass anything for the `$objDca` parameter or specifically passes `null`, then the `attributes_callback` and `options_callback` will receive `null` for their parameter.

leofeyer:
Hm, you are right.

PR #1783, opened 2026-07-31
Fix PHP version in DDEV setup

agonyz:
## Description

Update the PHP version in the DDEV installation examples from PHP `8.2` to PHP `8.4`.

The guide inserts the current Contao version into the installation command, while the PHP version in the preceding DDEV configuration is hard-coded. This can cause the two requirements to become inconsistent when the minimum PHP version changes.

This is currently the case for Contao 5.7 (which is shown as the default setup guide):

```console
ddev config --project-type=php --docroot=public --webserver-type=apache-fpm --php-version=8.2
ddev composer create-project contao/managed-edition:5.7
```

Following these commands in a clean project fails during dependency resolution because Contao 5.7 requires PHP `8.3` or higher.

Updating the DDEV configuration to PHP `8.4` ensures that the documented default setup works with the current Contao version shown on the same page.

### DIFF
--- a/docs/manual/guides/local-installation/ddev.de.md
+++ b/docs/manual/guides/local-installation/ddev.de.md
@@ -47,7 +47,7 @@ mkdir contao && cd contao
 DDEV-Konfiguration anlegen mit:
 
 ```shell
-ddev config --project-type=php --docroot=public --webserver-type=apache-fpm --php-version=8.3
+ddev config --project-type=php --docroot=public --webserver-type=apache-fpm --php-version=8.4
 ```
 
 Contao {{% siteparam "currentContaoVersion" %}} installieren:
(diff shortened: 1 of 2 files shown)


agonyz:
Another possible approach would be to introduce a PHP-version mapping for the supported Contao versions. However, I am not sure whether the additional complexity would be justified, or whether it is sufficient to ensure that the default setup works with the current Contao version by updating the PHP version in this guide.

A simpler alternative could be to introduce a separate site parameter, such as `recommendedPhpVersion`, alongside `currentContaoVersion`. This would keep both values in one central place and make it easier to update them together in the future.

fritzmg (approved):

fritzmg:
> Another possible approach would be to introduce a PHP-version mapping for the supported Contao versions.

That's not really necessary, unless you want to run legacy versions of Contao - which I think would be out of scope for this guide anyway.

PR #1784, opened 2026-08-17
Update contao version in example snippets

zonky2:


zoglo (approved):
Should be updated or kept general imo but unsure about the `^5.3 || ^6.0` constraint here. `^5.3` is an old LTS and whoever starts developing NEW extensions may want to start with `^5.7`. And not every extension will work with `^6.0` right away, we'd have to remove all `.html5` references from the docs... or rather, that should be the case for the `6.x` branch of the docs maybe.

fritzmg (approved):
Let's just use the `currentContaoVersion` parameter.

PR #1786, opened 2026-09-13
Mark the XAMPP documentation as outdated

zoglo:
### Description

This updates the documentation to reflect the current state of the XAMPP project.
grafik
grafik

XAMPP has been calling for contributors for a while and may receive Pull requests, there however weren't any releases anymore:
https://github.com/ApacheFriends/xampp-build/issues/47
https://mariadb.org/xampp-needs-more-apache-friends/

The last release has been made end of 2023:
https://www.apachefriends.org/blog/new_xampp_20231119.html

As for dependencies, the last PHP version it shipped with was PHP 8.2.12 (Security support for 8.2 ends end of this year - 2026 https://www.php.net/supported-versions.php)

To better reflect the state, I propose to add *warning* to the guide as the Xampp version does not allow you to even install Contao 5.7+ (Minimum required PHP Version is 8.3).

___

For reference:
https://community.contao.org/de/showthread.php?89454-Leidiges-Thema-contao-unter-Xampp&p=604708

### DIFF
--- a/docs/manual/guides/local-installation/xampp.de.md
+++ b/docs/manual/guides/local-installation/xampp.de.md
@@ -1,5 +1,5 @@
 ---
-title: "XAMPP"
+title: "XAMPP (veraltet)P"
 menuTitle : "Mit XAMPP"
 description: "Lokale Contao Installation mit XAMPP"
 weight: 30
@@ -9,6 +9,10 @@ tags:
    - "Installation"
 ---
 
+{{% notice warning %}}
+XAMPP wird seit längerer Zeit nicht mehr aktiv gepflegt und erlaubt keine Installation von neuen Contao-Versionen.
+Es sollte auf Alternativen ausgewichen werden.
+{{% /notice %}}
 
 In diesem Tutorial wird die lokale Contao Nutzung mit [XAMPP](https://www.apachefriends.org/) für Windows beschrieben. 
 Wir verwenden hierbei eine »XAMPP Portable Version«, die lediglich kopiert werden muß. Du wählst hierzu das 
--- a/docs/manual/guides/local-installation/xampp.en.md
+++ b/docs/manual/guides/local-installation/xampp.en.md
@@ -1,5 +1,5 @@
 ---
-title: "XAMPP"
+title: "XAMPP (outmoded)"
 menuTitle : "With XAMPP"
 description: "Contao installation with XAMPP"
 weight: 30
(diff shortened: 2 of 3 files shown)


docs/manual/guides/local-installation/xampp.de.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "XAMPP"
+title: "XAMPP (veraltet)P"
```
fritzmg:
```suggestion
title: "XAMPP (veraltet)"
```

docs/manual/guides/local-installation/xampp.en.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "XAMPP"
+title: "XAMPP (outmoded)"
```
fritzmg: I find "outmoded" very unusual. I think we should stick with common terminology, like

```suggestion
title: "XAMPP (deprecated)"
```

fritzmg (changes requested):

docs/manual/guides/local-installation/xampp.de.md
zoglo: Whoops :D

docs/manual/guides/local-installation/xampp.en.md
zoglo: I didn't want to use `deprecated` as a word since it hasn't been archived. Could use `outdated` but that would still not make the message clear 🤔 I guess `deprecated` could be used mhm

fritzmg: I think "out of date" would be a more suitable, technical term in this case. I have never seen "outmoded" anywhere 😄

zoglo: More denglish than anything 😁. Should we use `out of date` or `outdated` then?

fritzmg: I am in favour of `outdated`.

fkaminski: The same may apply to the [Devilbox](https://github.com/cytopia/devilbox), in which case it should be marked similarly :-).

zoglo: > The same may apply to the [Devilbox](https://github.com/cytopia/devilbox), in which case it should be marked similarly :-).

Might wanna open another PR for that 😁

zoglo: See 19f6b1cec33bc2635b459b932d418ef71d66c978 @fritzmg

PR #1787, opened 2026-09-17
Clarify Simple Tokens usage in Notification Center example

akroii:
Updated example for using Simple Tokens in the Notification Center to clarify usage with input fields.


zoglo:
Should we just redirect to the notification center docs instead?

fritzmg:
> Should we just redirect to the notification center docs instead?

Not really, as the functionality itself comes from Contao. We could however update the example to e.g. the confirmation text of forms, or the opt-in text for the registration module etc.

PR #1568, opened 2025-04-14
Add info on `primary` option for DCA operations

ameotoko:

PR #1788, opened 2026-09-20
Add primary attribute to button options

Tastaturberuf:


zoglo (approved):
Might be worth to mention that this is the same as #1568 but for global operations :)

fritzmg (approved):

